PR #21, opened 2019-11-05
2 changes to support development in PyCharm and debugging

jguertin:
- Added lines to the main app to run as a python script so that it can be run in the IDE
- Add an ignore line for PyCharm project directory

### DIFF
--- a/ppr-api/app/main.py
+++ b/ppr-api/app/main.py
@@ -1,3 +1,4 @@
+import uvicorn
 from fastapi import FastAPI
 from starlette.middleware.cors import CORSMiddleware
 
@@ -16,6 +17,11 @@ app.add_middleware(
     allow_headers=["*"]
 )
 
+
 @app.get("/")
 async def root():
     return {"message": "Hello World"}
+
+
+if __name__ == "__main__":
+    uvicorn.run(app, host="0.0.0.0", port=8000)
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ __pycache__
 
 # IDE
 /**/.vscode
+/.idea


ppr-api/app/main.py
```diff
@@ -16,6 +17,11 @@
     allow_headers=["*"]
 )
 
+
```
jguertin, line 20: Code inspection also caught this missing line

gh2os (approved):

ppr-api/app/main.py
WalterMoar: So this is what I'm afraid of when it comes to using a non-standard IDE (VS Code being our standard). Rather than try to get both IDEs to be happy, can we just stick to VS Code, at least until we get into a state where we can look at supporting both?

jguertin: I hear what you're saying.  But in this case I think it's a shortfall of pyLint, as it doesn't capture missing lines as it should.  If we integrate flake8 into our build pipeline, I think this mistake would be caught, as it is part of the PEP-8 style guide.

ppr-api/app/main.py
```diff
@@ -16,6 +17,11 @@
...
 @app.get("/")
 async def root():
     return {"message": "Hello World"}
+
+
+if __name__ == "__main__":
+    uvicorn.run(app, host="0.0.0.0", port=8000)
```
WalterMoar, line 27: Since we should be starting with "uvicorn app.main:app --reload", does this have any value? We'll want to have the "reload" autodeploys turned on when developing.

ppr-api/app/main.py
WalterMoar: OK, will look at getting Flake8 into VS Code.

ppr-api/app/main.py
jguertin: This was the recommended approach to enable debugging.  This only gets called if you launch the script directly.

https://fastapi.tiangolo.com/tutorial/debugging/

ppr-api/app/main.py
WalterMoar: OK, understood!

ppr-api/app/main.py
WalterMoar: Will revisit when we actually start using this API.

WalterMoar (approved):

PR #33, opened 2019-11-19
Build ims image

jguertin:
Created a Github Action workflow capable of generating a Docker image for the ims-api.  Currently, this does not do anything with the image besides building it, but this work will be added upon in the future.

Once we have full access to Artifactory, we can begin pushing the images and deploying to OpenShift.


bryan-gilbert:
Noticed something we may wish to add sometime. 

To shallow clone your repository, or copy only the latest version of your repository, set the fetch-depth with the with syntax:

- uses: actions/checkout@v1
  with:
    fetch-depth: 1

WalterMoar (approved):

PR #34, opened 2019-11-21
Web client (hello world) with openshift containerization

bryan-gilbert:
Basic containerization.  Two build images and one Deploy config.  First build the web Vue static pages. Then build an image with a Caddy server that provides these static files.  The deploy config includes a service and route that serves the Vue client on a public url.


.gitignore
```diff
@@ -5,10 +5,32 @@ __pycache__
...
 # Environments
 .venv
+.env.local
+.env.*.local
 
-# IDE
-/**/.vscode
-/**/.idea
+
+.DS_Store
```
jguertin: `.DS_Store` has a duplicate entry

jguertin (commented):
Looks good to me.  Comments were discussed in person.

ppr-client/openshift/ppr-ui-bc.tools.yaml
```diff
@@ -0,0 +1,68 @@
...
+    output:
+      to:
+        kind: ImageStreamTag
+        name: ${APP_NAME}:latest
+    source:
+      dockerfile: |
+        FROM bcgov-s2i-caddy
+        RUN mkdir -p /var/www/html
+        COPY dist /var/www/html
+        RUN ls -alR /var/www/html
```
WalterMoar: This final RUN command was for debugging. It should be removed.

ppr-client/openshift/ppr-ui-bc.tools.yaml
```diff
@@ -0,0 +1,68 @@
...
+- description: The name of the appliction
+  displayName: APP Name
+  name: APP_NAME
+  required: true
+  value: ppr-ui
+- description: The path at which web application is deployed. Context root for the web applicaton
+  displayName: WEB_APP_CONTEXT_PATH
+  name: WEB_APP_CONTEXT_PATH
+  required: true
+  value: ppr
```
WalterMoar, line 67: This doesn't appear to be used in this bc - can you remove it?

ppr-client/openshift/ppr-ui-dc.yaml
```diff
@@ -0,0 +1,170 @@
...
+- description: The path at which web application is deployed. Context root for the web applicaton
+  displayName: WEB_APP_CONTEXT_PATH
+  name: WEB_APP_CONTEXT_PATH
+  required: true
+  value: ppr
+- description: The port for the web application
+  displayName: Web application port
+  name: APP_PORT
+  required: true
+  value: "2015" #See notes above where this can not be used
```
WalterMoar, line 160: I think we're OK hard-coding the port for the dc. Since every pod has its own IP address, we will never have two deploys in the same pod, so no worries about port contention.

While we're on the topic: can we run it on port 8080? It would be nice to have all web servers running on the same port, to reduce "cognitive load".

ppr-client/openshift/ppr-ui-inter-bc.tools.yaml
```diff
@@ -0,0 +1,47 @@
...
+        kind: ImageStreamTag
+        name: ${APP_NAME}:latest
+    source:
+      contextDir: ppr-client
+      dockerfile: |
+        FROM node:8
+        COPY . .
+        RUN npm install
+        RUN npm run build
+        RUN ls -alR /dist
```
WalterMoar: This looks like another debugging statement that can be removed.

ppr-client/openshift/ppr-ui-dc.yaml
```diff
@@ -0,0 +1,170 @@
...
+- description: The project's openshift DEV (deploy) namespace
+  displayName: Dev namespace
+  name: NS_DEV
+  required: true
+  value: zwmtib-dev
+- description: The project's openshift TOOLS (build) namespace
+  displayName: TOOLS namespace
+  name: NS_TOOLS
+  required: true
+  value: zwmtib-tools
```
WalterMoar: GitHub is complaining about no newline at the end, so I guess it can't hurt to put it in.

ppr-client/openshift/ppr-ui-dc.yaml
bryan-gilbert: Actually, the default Caddy port is 2015.  We can change this once we add in configuration and volume mounts.  That's a time when having this one-stop location for port will help.

vcli/README.md
```diff
@@ -0,0 +1,57 @@
...
+to create a Vue Hello World project in a new folder called ```my-client```
+
+
+```shell script
+node  node_modules/@vue/cli/bin/vue.js create ppr-client
+```
+Follow the prompts. There is an option to select a manual set of options. For the initial POC we selected all options available. Including:
+   - Babel
+   - TypeScript
+   - Progressive Web App (PWA) Support
```
WalterMoar, line 24: Just to confirm - we're building a SPA that is also a PWA?

WalterMoar (changes requested):
The big one is that for consistency we probably want this all to be under /ppr-ui, rather than /ppr-client.

ppr-client/openshift/ppr-ui-dc.yaml
WalterMoar: OK, looks like all the projects are going with 2015 for any caddy deploys. Since we also want to be consistent with the other teams let's stick with 2015. I think it's better to hard-code for everything, if we have to hard-code for some.

WalterMoar:
As discussed everything is good for the first pull.

WalterMoar (approved):

PR #37, opened 2019-11-22
Apply configuration to support health-check and environment variables

jguertin:
Add secret environment variable references and health check to the deployment config.  Also corrected the IMS dataStore configurable value so it can support environment variables.

### DIFF
--- a/ims-api/openshift/ims-api-dc.yaml
+++ b/ims-api/openshift/ims-api-dc.yaml
@@ -30,11 +30,25 @@ objects:
           deploymentconfig: ${API_NAME}
       spec:
         containers:
-        - imagePullPolicy: Always
+        - envFrom:
+            - prefix: IMS_
+              secretRef:
+                name: ims-connect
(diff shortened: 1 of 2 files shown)


ims-api/openshift/ims-api-dc.yaml
```diff
@@ -30,11 +30,25 @@ objects:
...
+          readinessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /actuator/health
+              port: 8080
+              scheme: HTTP
+            initialDelaySeconds: 1
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 10
```
WalterMoar: There's a liveness probe too - for the most part everyone has them doing essentially the same thing. I think it's something that needs a whole bunch of investigation, but do be aware that there is the other one too. (don't think we need it now, though, unless OpenShift complains about it).

WalterMoar (approved):

PR #38, opened 2019-11-22
Sprint 0 Vue client work

bryan-gilbert:
Move the docs about how to create and update the client into the client folder.
Remove the gitignore file that was generated by the VUE cli. (Depend on the root gitignore)
Fix a unit test.
Add a package.json script to run both the unit and e2e tests.
Typescript use strict mode
Client - add pug, vuetify, do audit fix
npm audit fix
npm install -D pug pug-plain-loader
npm install vuetify
Switch from less to sass/scss to match lear
npm uninstall less
npm uninstall less-loader
npm install -D sass
npm install -D sass-loader
To explain the next install see the sass-loader docs https://www.npmjs.com/package/sass-loader
npm install --save-dev fibers

Import initial styles from lear

Remove generated Vue
Add SBC common headers and footers
npm install sbc-common-components --save

Axios, Configuration, Routes, Auth stub
npm install --save axios

Prefetch configuration
Router and routes
Route hook to check for authorization and redirect if needed

Public vs User page layout templates
BC Reg icons
Initial editor configuration

Create sample dashboard
Create a sample Home page
Create a fake login page
Flow and routing between these.
consistent home and dashboard page layout
code format


ppr-client/src/App.vue
```diff
@@ -1,32 +1,63 @@
...
+                }.bind(this), delay)
+            })
+        },
+        errorCaptured(err, vm, info) {
+            console.log('App errorCaptured', err)
+            // err: error trace
+            // vm: component in which error occured
+            // info: Vue specific error information such as lifecycle hooks, events etc.
+            // TODO: Perform any custom logic or log to server
+            // return false to stop the propagation of errors further to parent or global error handler
```
WalterMoar, line 49: We will eventually look at sentry.io here for logging that we can see.

ppr-client/package.json
```diff
@@ -5,17 +5,21 @@
   "scripts": {
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
+    "test": "npm run test:e2e && npm run test:e2e",
```
jguertin: Seems to be running the tests twice

ppr-client/public/config/configuration.json
```diff
@@ -0,0 +1,7 @@
+{
+  "API_URL": "https://WHAT_TO_PUT_HERE.pathfinder.gov.bc.ca/api/v1/",
+  "AUTH_URL": "https://dev.bcregistry.ca/WHAT-TO_PUT_HERE/auth/",
+  "AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/",
+  "PAY_API_URL": "https://pay-api-dev.pathfinder.gov.bc.ca/api/v1/",
+  "ADDRESS_COMPLETE_KEY": "RH99-RN99-GB72-FW86"
```
jguertin, line 6: Should these details be in the source code?

ppr-client/public/config/configuration.json
```diff
@@ -0,0 +1,7 @@
...
+}
```
WalterMoar, line 7: Wondering if we want to keep this file and toss the others. It'll be overwritten by a configmap.

ppr-client/.env
```diff
@@ -0,0 +1 @@
+VUE_APP_PATH = ppr
```
WalterMoar, line 1: Save to assume that this file will disappear? We don't usually check in the .env files.

ppr-client/src/App.vue
```diff
@@ -1,32 +1,63 @@
...
+            }
+        },
+        mounted() {
+            console.log('App mounted')
+            const delay = 1111
+            this.$nextTick(() => {
+                setTimeout(function () {
+                    this.setLoaded(true)
+                }.bind(this), delay)
+            })
```
WalterMoar: This looks fragile - assume it's from LEAR?

ppr-client/.env
bryan-gilbert: yes

ppr-client/public/config/configuration.json
bryan-gilbert: will remove address key

ppr-client/src/App.vue
bryan-gilbert: no -- left over early dev.  Will remove this in next PR

ppr-client/package.json
bryan-gilbert: Good catch. Will fix in next PR

ppr-client/src/assets/styles/base.scss
```diff
@@ -0,0 +1,278 @@
...
+  html {
+    font-size: 16px;
+  }
+}
+
+body {
+  // NB: Vuetify font is set on .v-application class (see below)
+  // font-family: 'BCSans', sans-serif;
+  font-family: 'Roboto', sans-serif;
+  font-size: 1rem;
```
WalterMoar: I know this is probably just copied from LEAR but I wonder why not either use the BC font here, or if it doesn't do anything then remove it? CSS is one of those things that is so hard to clean up after the fact :)

ppr-client/src/assets/styles/base.scss
```diff
@@ -0,0 +1,278 @@
...
+  margin-top: 0.25rem;
+  color: $gray6;
+  font-size: 0.875rem;
+  font-weight: 500;
+}
+
+.v-application {
+  // override the default Vuetify font
+  // font-family: 'BCSans', sans-serif;
+  font-family: 'Roboto', sans-serif;
```
WalterMoar: Should be the BC font here, yes?

ppr-client/src/assets/styles/base.scss
bryan-gilbert: This is copied from LEAR.  Can we leave this to a future "sync UI styles with LEAR" user story.

ppr-client/src/registerServiceWorker.ts
```diff
@@ -1,31 +1,31 @@
 /* eslint-disable no-console */
 
-import { register } from 'register-service-worker'
+import {register} from 'register-service-worker'
```
WalterMoar, line 3: I'm a bit curious about the formatting. We'll want to use whatever is the default for VS Code. We'll figure that out as we go, though.

ppr-client/src/utils/config-helper.ts
```diff
@@ -0,0 +1,41 @@
+import axios from '@/utils/axios-auth'
+
+export default {
+  /**
+   * fetch config from API
+   */
+  fetchConfig() {
+    // const url = `/${process.env.VUE_APP_PATH}/config/stub-configuration.json`
```
WalterMoar, line 8: I'm fairly keen on not committing commented-out code, but let's just get this monster merged.

ppr-client/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "esnext",
     "module": "esnext",
-    "strict": true,
+    "strict": false,
```
WalterMoar: This is the only one of real concern for me. If we turn it off now, it will get harder and harder to turn it on.

WalterMoar (changes requested):
The only change I'd like to see is the TS compiler strict flag. Disabling the checking will cause too much rework with no business benefit, and that's a hard sell.

ppr-client/src/registerServiceWorker.ts
bryan-gilbert: Me too.  Seeking standardization with automation.

ppr-client/src/assets/styles/base.scss
WalterMoar: I'd like to be careful about building up technical debt. I can't see that kind of thing being approved when there are missing features.

ppr-client/tsconfig.json
bryan-gilbert: This setting was copied from LEAR.  I agree with you and think we should change this back (in the next PR).

WalterMoar (approved):

PR #40, opened 2019-11-25
Add faked feature flags, toggles access to faked login

bryan-gilbert:


ppr-client/src/App.vue
```diff
@@ -2,18 +2,24 @@
...
-        components: {},
+        components: { FeatureOne, FeatureTwo },
 
         data: function () {
             return {
-                dataLoaded: false
+                dataLoaded: false,
             }
         },
         provide() {
```
WalterMoar, line 25: We should start to think about using the TypeScript class components instead of the javascript-style components. The co-ops has some good examples.

WalterMoar (approved):

ppr-client/src/App.vue
bryan-gilbert: Absolutely.  Look for a PR coming our way soon.

PR #48, opened 2019-11-30
UI makes web call for a faked out auth

bryan-gilbert:
FeatureOne enables the display of a "log on" button.
This takes the user to a "log in" screen.
Enter the user id and log on
Calls to PPR API to authenticate.
On success take user to a success page. Let user go to dashboard.
All of the above is just to prove concepts and will be replaced piece by piece as we connect to the real authentication servers, and launch darkly for feature flag management.

As well, this PR fixes some issues with the BC and DC files for the PPR-API


jguertin (approved):

bryan-gilbert:
replaced with new PR https://github.com/bcgov/ppr/pull/49

PR #49, opened 2019-12-03
UI auth mockup - Vue Composition API - 

bryan-gilbert:
Setup the configuration file.
Fetch config on page load using a workaround until we have a web path like lear
Load a application configuration object.
Provide a Vue component to display configuration, for dev only
Configuration, web api call from client,  api fakes auth
set build config to pull from my repo for testing
Fix round trip, use auth api url
Remove service worker pending further research
Fix ppr-api bc and dc scripts
Refactor to leverage the composition API, which will be part of Vue core in their upcoming 3.0 release.
Removed references to the Component decorator in favour of `createComponent` and added explicit type to arrow function in the Vue components.
Adjust views and components to new Vue Composition API

### DIFF
--- a/docs/ppr-api/development.md
+++ b/docs/ppr-api/development.md
@@ -33,8 +33,7 @@ This will install what is needed to run in VS Code. Note that there will be erro
 ## Start the Web Service
 
 ```
-(.venv) $ cd src
-(.venv) $ uvicorn main:app --reload
+(.venv) $ uvicorn src.main:app --reload
 ```
 
 This should bring the web service up on http://localhost:8000, with OpenAPI documentation available at http://localhost:8000/docs.
--- a/docs/ppr-api/development.md
+++ b/docs/ppr-api/development.md
@@ -16,10 +16,16 @@ Popup: "Linter pylint is not installed." (`Install`)
 
 Ensure the command line prompt starts with something like `(.venv)`, otherwise ... (TBD: how to activate?).
 
+Or from the bash command line activate the virtual environment
+
+```shell script
+source ./.venv/bin/activate
+```
+
 ## Installing Development Requirements
 
 ```
-$ pip3 install -r requirements/dev.txt
+(.venv) $ pip3 install -r requirements/dev.txt
 ```
 
 This will install what is needed to run in VS Code. Note that there will be errors regarding an invalid command 'bdist_wheel', but these don't appear to be a problem.
@@ -27,7 +33,8 @@ This will install what is needed to run in VS Code. Note that there will be erro
 ## Start the Web Service
 
 ```
-$ uvicorn app.main:app --reload
+(.venv) $ cd src
+(.venv) $ uvicorn main:app --reload
 ```
 
 This should bring the web service up on http://localhost:8000, with OpenAPI documentation available at http://localhost:8000/docs.
--- a/ppr-api/openshift/ppr-api-dc.yaml
+++ b/ppr-api/openshift/ppr-api-dc.yaml
@@ -32,13 +32,13 @@ objects:
         containers:
         - env:
           - name: PORT
-            value: "8000"
+            value: "8080"
           - name: WEB_CONCURRENCY
             value: "4"
           imagePullPolicy: Always
           name: ${API_NAME}
           ports:
-          - containerPort: 8000
+          - containerPort: 8080
             protocol: TCP
           resources:
             limits:
@@ -74,10 +74,10 @@ objects:
     name: ${API_NAME}
   spec:
     ports:
-    - name: 8000-tcp
-      port: 8000
+    - name: 8080-tcp
+      port: 8080
       protocol: TCP
-      targetPort: 8000
+      targetPort: 8080
     selector:
       deploymentconfig: ${API_NAME}
     sessionAffinity: None
@@ -91,7 +91,7 @@ objects:
   spec:
     host: ${ROUTE_URL}
     port:
-      targetPort: 8000-tcp
+      targetPort: 8080-tcp
     tls:
       insecureEdgeTerminationPolicy: Redirect
       termination: edge
--- a/ppr-api/openshift/ppr-api-dc.yaml
+++ b/ppr-api/openshift/ppr-api-dc.yaml
@@ -32,13 +32,13 @@ objects:
         containers:
         - env:
           - name: PORT
-            value: "8080"
+            value: "8000"
           - name: WEB_CONCURRENCY
             value: "4"
           imagePullPolicy: Always
           name: ${API_NAME}
           ports:
-          - containerPort: 8080
+          - containerPort: 8000
             protocol: TCP
           resources:
             limits:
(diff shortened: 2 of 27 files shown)


ppr-ui/src/App.vue
```diff
@@ -2,59 +2,64 @@
...
+    components: { FeatureOne, FeatureTwo },
+    setup(): Data {
+      provideRouter()
+      const router = useRouter()
+
+      provide("originUrl", origin())
+      provide("authApiUrl", authAPIURL())
+      provide("featureOne", ref(AppData.features.featureOne))
+      provide("featureTwo", ref(AppData.features.featureTwo))
+      provide("configuration", ref(AppData.config))
```
jguertin, line 42: Should `"configuration"` be `reactive` rather than `ref`?  I believe `ref` is intended to wrap primitives.

ppr-ui/src/components/ConfigInfo.vue
```diff
@@ -0,0 +1,42 @@
...
+    import {Data} from "@vue/composition-api/dist/ts-api/component";
+
+    export default createComponent({
+        setup(): Data {
+            const configuration = inject("configuration", reactive({}))
+            return {configuration}
+        }
+    })
+</script>
+<style lang="scss">
```
jguertin, line 34: style should be `scoped` for components. https://vuejs.org/v2/style-guide/#Component-style-scoping-essential

ppr-ui/src/components/ConfigInfo.vue
```diff
@@ -0,0 +1,42 @@
...
+  dt {
```
jguertin, line 35: We should favour class selectors over element selectors: https://vuejs.org/v2/style-guide/#Element-selectors-with-scoped-use-with-caution

Granted this is a low priority recommendation, and atm it isn't scoped, so ignore this if you will.

ppr-ui/src/layouts/LayoutPublic.vue
```diff
@@ -10,21 +10,17 @@
...
-        inject: ['authApiUrl', 'originUrl'],
-    }
+  export default createComponent({
+    name: 'LayoutPublic',
+    components: {
+      SbcHeader,
+      SbcFooter,
+    },
+    inject: ['authApiUrl', 'originUrl']
+  })
```
jguertin, line 25: Just for @WalterMoar's benefit, one of the nice things about the Composition API is that the old way of doing things still works.  In this case, it was a better not to use `setup()` since the only thing to provide are the injects, which would have been overkill in this case.

ppr-ui/src/utils/app-data.ts
```diff
@@ -1,30 +1,110 @@
...
+    sessionStorage.setItem('FEATURE_TWO', flag ? 'true' : 'false')
     this._featureTwo = flag
   }
 }
 
+/*
+Config
+Hides the implementation of storing configuration information.
+ */
+class Config {
```
jguertin, line 27: Would this be better as an immutable class?  Is there ever a need to mutate the content after initialization?

To take it a bit further, are the values loaded from a configuration file?  If so, is there any need to use session storage at all?

ppr-ui/src/App.vue
bryan-gilbert: Perhaps but since the underlying object is a primative ref seems more appropriate.  There is yet o be clear guidance on when to use ref vs reactive so for now shall we (suggestion) keep it simple and use ref for primatives?

ppr-ui/src/components/ConfigInfo.vue
bryan-gilbert: Yes.  Normal practice.  Not so important in this throw away component.

ppr-ui/src/components/ConfigInfo.vue
bryan-gilbert: Absolutely except when the component is a throw-away.

ppr-ui/src/utils/app-data.ts
bryan-gilbert: Two good points.  We can make this immutable. And there is no need for session storage once the values are loaded into memory because the settings do come from a GET when the app is loaded.  Suggestion that we make this change in a follow up PR

docs/ppr-api/development.md
```diff
@@ -16,18 +16,25 @@ Popup: "Linter pylint is not installed." (`Install`)
...
 ```
 
 This will install what is needed to run in VS Code. Note that there will be errors regarding an invalid command 'bdist_wheel', but these don't appear to be a problem.
 
 ## Start the Web Service
 
 ```
-$ uvicorn app.main:app --reload
+(.venv) $ cd src
+(.venv) $ uvicorn main:app --reload
```
WalterMoar: I see what's going on here now. The 'cd' can be removed and the line should be:

> (.venv) $ uvicorn src.main:app --reload

ppr-api/openshift/ppr-api-dc.yaml
```diff
@@ -32,13 +32,13 @@ objects:
         containers:
         - env:
           - name: PORT
-            value: "8080"
+            value: "8000"
```
WalterMoar: Please revert this, there's no need to change to 8000.

ppr-api/src/main.py
```diff
@@ -23,5 +23,18 @@
...
+@app.get("/auth/{user_name}", status_code=201)
+async def read_item(user_name: str, user_agent: str = Header(None)):
+    my_list = ['justin', 'walter', 'bryan']
+    print(my_list)
+    print(user_name)
+    if user_name in my_list:
+        return {"user_name": user_name, "user_agent": user_agent}
+    raise HTTPException(status_code=404, detail="Invalid user name")
+
+
```
WalterMoar, line 38: Please revert this - we should avoid putting time into throwaway code.

WalterMoar (changes requested):
Let's discuss - PRs should be small and focused on a single task.

ppr-ui/src/utils/app-data.ts
jguertin: Agreed

ppr-api/openshift/ppr-api-dc.yaml
bryan-gilbert: In main.py the port is 8000 and this change allows the container to serve that port.

docs/ppr-api/development.md
bryan-gilbert: thanks .. fixed

ppr-api/openshift/ppr-api-dc.yaml
bryan-gilbert: backed out

ppr-api/src/main.py
bryan-gilbert: as per conversation will remove this in next PR

ppr-ui/openshift/ppr-ui-dc.yaml
```diff
@@ -64,11 +64,20 @@ objects:
...
+            volumeMounts:
+              - mountPath: /var/www/html/config
+                name: ppr-web-ui-config
+                readOnly: true
         dnsPolicy: ClusterFirst
         restartPolicy: Always
         schedulerName: default-scheduler
         securityContext: {}
-        terminationGracePeriodSeconds: 30
+        terminationGracePeriodSeconds:
```
WalterMoar: The setting here seems to have been lost.

WalterMoar (changes requested):
PR was too big to be able to do a good review. I did catch that one dc had a lost parameter, so please replace that and it should be good.

WalterMoar (approved):

PR #55, opened 2019-12-04
Set up the OpenShift labels for bc and dc. #65

WalterMoar:
Updated the buildconfig and deployconfig to have labels than can be used in the networksecuritypolicy work.

### DIFF
--- a/ims-api/openshift/ims-api-bc.yaml
+++ b/ims-api/openshift/ims-api-bc.yaml
@@ -6,7 +6,9 @@ objects:
   kind: BuildConfig
   metadata:
     labels:
+      app: ppr
       build: ${API_NAME}
+      role: ${API_NAME}
     name: ${API_NAME}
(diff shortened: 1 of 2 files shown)


ims-api/openshift/ims-api-bc.yaml
```diff
@@ -30,14 +32,14 @@ objects:
         uri: https://github.com/bcgov/ppr
       type: Git
     strategy:
-      dockerStrategy:
-        noCache: false
+      dockerStrategy: {}
```
WalterMoar, line 35: The noCache default value is false, but having this in the bc was causing the "apply" to always say "configured" when repeatedly applied (expect to be "unchanged").

jguertin (approved):

bryan-gilbert (approved):
Makes sense ... getting ready for the zero trust network.

PR #59, opened 2019-12-05
Support stubbing data in ims-api

jguertin:
Refactored the implementation to stub out search functionality without needing to interact with the IMS Server

My goal with this commit is to extract the IMS Connect portion of the application so we can begin stubbing out the API without being blocked by having the interactivity functioning.


WalterMoar (approved):
Niiiice.

bryan-gilbert:
Nice unblocker ... what is the api call that reaches this stub?

jguertin:
> Nice unblocker ... what is the api call that reaches this stub?

atm, the only path is to the stub.  You can change it locally by modifying the qualifier in the controller.

PR #61, opened 2019-12-05
Adjusted health check delays and timeouts. #73

WalterMoar:
There were errors in OpenShift due to the readiness check being called before the pods had time to spin up. Adjusted the delays to give the pods some time to start. Opted to go with the best-case time, so that the rolling deploys go faster, but note that if the cluster is busy then we will get errors (but the checks will retry three times before giving up).


ppr-ui/openshift/ppr-ui-dc.yaml
```diff
@@ -58,9 +58,10 @@ objects:
                 path: /
                 port: 2015
                 scheme: HTTP
+              initialDelaySeconds: 5
               periodSeconds: 10
               successThreshold: 1
-              timeoutSeconds: 1
+              timeoutSeconds: 10
```
WalterMoar, line 64: This was a bit strict.

jguertin (approved):

bryan-gilbert (approved):

PR #62, opened 2019-12-06
Rework feature flag internals

bryan-gilbert:
In preparation of using Launch Darkly refactor the way we manage and use feature flags in the code. Note these flags are reactive.
Next steps: remove the components that set feature flags and, instead, get the values from LD.

### DIFF
--- a/ppr-ui/src/components/FeatureOne.vue
+++ b/ppr-ui/src/components/FeatureOne.vue
@@ -1,31 +1,33 @@
 <template>
   <div class="features">
-    <v-checkbox class="f-check" id="featureOne" v-model="featureOneFlag" :label="fOneToggleLabel" />
+    <v-checkbox class="f-check" v-model="featureFlag" :label="featureLabel" />
   </div>
 </template>
 
 <script lang="ts">
-import {computed, createComponent, inject, ref, watch} from "@vue/composition-api"
+import {computed, createComponent, watch} from "@vue/composition-api"
 import {Data} from "@vue/composition-api/dist/component"
-import AppData from '@/utils/app-data'
+import {useFlags} from '@/flags/feature-flags'
 
 export default createComponent({
   setup(): Data {
-    const featureOneFlag = inject("featureOne", ref(false))
-    const fOneToggleLabel = computed((): string => featureOneFlag.value ? 'Disable F One' : ' Enable F One')
+    const name = 'One'
+    const flags = useFlags()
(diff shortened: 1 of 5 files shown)


ppr-ui/src/flags/feature-flags.ts
```diff
@@ -0,0 +1,32 @@
+import {ref, provide, inject, Ref} from "@vue/composition-api"
+
+// const ldKeys = ['poc-feature-1', 'poc-feature-2']
+
+export class FeatureFlags {
+  pocFeature1: Ref<boolean>
+  pocFeature2: Ref<boolean>
```
jguertin, line 7: I don't think we should be using `Ref`'s at this level.  Mixing the Vue framework into the data-types breaks the single responsibility principal.  Rather, I think we should be using `reactive` around instances of FeatureFlags, if we're looking to capture mutations.

ppr-ui/src/views/Home.vue
```diff
@@ -6,6 +6,14 @@
           <h1>PPR Sample Home Page</h1>
         </header>
 
+        <div>
+          <h4>Feature Flags</h4>
+          <ul>
+            <li>poc 1 {{featureOne}}</li>
+            <li>poc 2 {{featureTwo}}</li>
```
jguertin, line 13: I think if `flags` (below) is reactive, you can use `flags.featureTwo` here and see changes updated.

ppr-ui/src/components/FeatureOne.vue
```diff
@@ -1,31 +1,33 @@
...
   setup(): Data {
-    const featureOneFlag = inject("featureOne", ref(false))
-    const fOneToggleLabel = computed((): string => featureOneFlag.value ? 'Disable F One' : ' Enable F One')
+    const name = 'One'
+    const flags = useFlags()
+    const featureFlag = flags.pocFeature1
+    const featureLabel = computed((): string => (featureFlag.value ? 'Disable' : ' Enable') + ' Feature ' + name)
 
-    watch(featureOneFlag, (flag): void => { AppData.features.featureOne = flag })
+    watch(featureFlag, (flag): void => { flags.pocFeature1.value = flag })
```
jguertin, line 19: I'm not a fan of components mutating state they don't own directly.  This should be centralized, ideally with some notification to merge state changes, but at minimum a setter that ensure controlled and safe change of state.

I've been burned by some really hard to diagnose bugs by not having very strict controls about changing state.

PR #64, opened 2019-12-06
Feature flags

bryan-gilbert:
- include lint fixes also submitted by Justin
In preparation of using Launch Darkly refactor the way we manage and use feature flags in the code. Note these flags are reactive.
Next steps: remove the components that set feature flags and, instead, get the values from LD.
Simplify approach


bryan-gilbert:
Now what?   -- hold on .. back soon

PR #65, opened 2019-12-09
Refactor Feature Flags for Reactivity

jguertin:
Combined work with Bryan to begin the process of integrating Launch Darkly.  This step refactors our existing placeholder FF approach to leverage reactivity.  By providing the "global" ff object in the App using `reactive()`, when we inject the object into components and leverage it in templates, the UI will respond immediately to changes in settings.

The following step after this commit will be to replace the internals with Launch Darkly integration.

### DIFF
--- a/ppr-ui/src/views/Home.vue
+++ b/ppr-ui/src/views/Home.vue
@@ -34,7 +34,7 @@
       </article>
     </v-container>
 
-    <v-container v-if="featureOne">
+    <v-container v-if="flags.feature1">
       <v-btn
         class="form-primary-btn"
         color="primary"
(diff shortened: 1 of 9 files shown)


bryan-gilbert (approved):

ppr-ui/src/views/Home.vue
```diff
@@ -35,29 +43,29 @@
...
 export default createComponent({
   setup(): Data {
     const router = useRouter()
-    const featureOne = inject("featureOne", ref(false))
-    const featureTwo = inject("featureTwo", ref(false))
+
+    // Feature Flags
+    const flags = useFeatureFlags()
+    const featureOneLabel = computed((): string => (flags.feature1 ? 'enabled' : ' disabled'))
+    const featureTwoLabel = computed((): string => (flags.feature2 ? 'enabled' : 'disabled'))
```
WalterMoar, line 62: Is the difference in the disabled strings intentional?

WalterMoar: Oh... is this a label that we'll throw away? If so then ignore the above.

WalterMoar (approved):
I haven't had time to look into the Composition API, so not sure if I can give this a good review without doing so. Will try to play with it this week, though.

ppr-ui/src/views/Home.vue
bryan-gilbert: Yes this is throw away

PR #66, opened 2019-12-09
Let Launch Darkly control the feature flags

bryan-gilbert:
To use MUST set the LD key in the config map in openshift.
Please apply PR #65 first.


bryan-gilbert:
in progress ... reworking a unit test

ppr-ui/public/config/sample-dev-configuration.json
```diff
@@ -4,5 +4,6 @@
   "AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/",
   "PAY_API_URL": "https://pay-api-dev.pathfinder.gov.bc.ca/api/v1/",
   "SENTRY_DSN": "",
-  "SENTRY_ENVIRONMENT": "dev"
+  "SENTRY_ENVIRONMENT": "dev",
+  "LAUNCH_DARKLY_CLIENT_KEY": ""
```
WalterMoar: Can we keep these alphabetical? It makes them easier to scan through to find something.

ppr-ui/src/App.vue
```diff
@@ -37,10 +31,23 @@ function authAPIURL(): string {
...
+/*
+  NOTE: the following userKey is TEMPORARY and will disappear when we get auth set up.
+  Create one user id per session.
+ */
+const userKey = sessionStorage.getItem('userKey') ? sessionStorage.getItem('userKey') : uuid.v4()
+sessionStorage.setItem('userKey', userKey)
+
 export default createComponent({
-  components: { FeatureOne, FeatureTwo },
+  components: { },
```
WalterMoar: Can this line be removed?

ppr-ui/src/flags/feature-flags.ts
```diff
@@ -4,11 +4,17 @@ export class FeatureFlags {
   private _feature1: boolean
   private _feature2: boolean
 
-  public constructor(feature1: boolean = false, feature2: boolean = false) {
+  private static _instance: FeatureFlags
+
+  private constructor(feature1: boolean = false, feature2: boolean = false) {
```
WalterMoar, line 9: I think this is good for now, but I can see it being a problem when we a lot of flags. Is VueX not an option?

bryan-gilbert: Exactly.  This was a design decision that we need to revisit.  The current assumption is we'll not have very many flags (under 10) in play at any time. We can pause and reconsider this assumption now or push that out to a later time.  Open to guidance here.

WalterMoar: It's definitely not something to worry about for this PR but something good to sort out soon. Wish I had guidance but let's keep pondering :)

ppr-ui/public/config/configuration.json
```diff
@@ -4,5 +4,6 @@
   "AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/",
   "PAY_API_URL": "https://pay-api-test.pathfinder.gov.bc.ca/api/v1/",
   "SENTRY_DSN": "",
-  "SENTRY_ENVIRONMENT": "local.developer"
+  "SENTRY_ENVIRONMENT": "local.developer",
+  "LAUNCH_DARKLY_CLIENT_KEY": "Client side ID from https://app.launchdarkly.com/settings/projects"
```
WalterMoar: I guess we should alphabetize these for consistency.

ppr-ui/src/flags/feature-flags.ts
jguertin: I would imagine we'll end up taking a data structure rather than a bunch of individual components, so I vote cross the bridge when we get there.

WalterMoar (changes requested):
Just the minor change to ppr-ui/public/config/configuration.json and then it's good to go.

ppr-ui/src/utils/config-helper.ts
```diff
@@ -36,21 +38,22 @@ export default {
       'ResponseType': 'application/json',
       'Cache-Control': 'no-cache'
     }
-    console.log('Fetch config from ', url)
+    if(debug) console.log('Fetch config from ', url)
```
jguertin, line 41: Generally speaking, we should avoid committing `console.log()`.  If you do want debugging to be always be performed in the client, you can use `console.debug()` (without `if (debug)`).  You can then adjust the logs you view with the in browser tools.

ppr-ui/src/flags/ld-user.ts
```diff
@@ -0,0 +1,50 @@
+import ip from 'ip'
+import UAParser from 'ua-parser-js'
+import {LDUser} from 'launchdarkly-js-client-sdk'
+
+const userAgentParser = new UAParser()
+
+function baseUser(): LDUser {
+  // return a LDUser ... defined in https://github.com/launchdarkly/js-sdk-common/blob/master/typings.d.ts
```
jguertin, line 8: I'm assuming commented code will be removed quickly?

jguertin (changes requested):

WalterMoar (approved):

PR #68, opened 2019-12-12
Clean up lint warnings and configuration

bryan-gilbert:
Add a precommit script to package.json
Add comments to package.json
Configure webpack to remove entry point size limits warnings.
Clean up configuration initialization (use immutable approach, sort members, etc)
Add TS types to remove lint warnings
Disable lazy load of components in routes until we an determine a function return type.


ppr-api/README.md
```diff
@@ -0,0 +1,44 @@
+# PPR API
```
WalterMoar: This file is going to cause merge issues with my ppr-api work.

ppr-ui/src/utils/auth-helper.ts
```diff
@@ -1,12 +1,14 @@
...
+
 export default {
   authRedirect(): void {
     const redirected = sessionStorage.getItem('REDIRECTED')
-    const authUrl: string = sessionStorage.getItem('AUTH_URL') || ''
-    // console.log('redirected contains', redirected)
-    // console.log('authUrl contains', authUrl)
+    const authUrl = AppData.config.authUrl
+    if(debug) console.log('AuthHelper redirected contains', redirected)
+    if(debug) console.log('AuthHelper authUrl contains', authUrl)
```
WalterMoar, line 11: I'm going to suggest that everyone use VS Code. If I touch any of the ppr-ui or ims-api files, VS Code is going to reformat them.

bryan-gilbert: I've installed and run VSCode, and I've opened the file mentioned above. I don't see it making any changes.   I'll switch to VSCode going forward yet I wonder if there is some settings I need to use to get the same as you do.  Also, consider that this is an open source project and other contributors need to be able to make submissions without prescribing a particular tool set.  These formatting rules need to be in our code base. IMHO

ppr-api/README.md
bryan-gilbert: I can back out this file.

jguertin (approved):

PR #70, opened 2019-12-12
Parse and return the results of a Search against IMS Connect

jguertin:
Created a class to parse search output and then used it to provide vehicles to the client through the existing API call.


bryan-gilbert (approved):

WalterMoar:
Ship it!

PR #73, opened 2019-12-13
Add a "Health Check" endpoint for readiness probes to call on OpenShift

jguertin:


bryan-gilbert (commented):
nice and simple

bryan-gilbert (approved):

ppr-api/src/endpoints/healthcheck.py
```diff
@@ -0,0 +1,8 @@
+from fastapi import APIRouter
+
+router = APIRouter()
```
WalterMoar: I'd like to go with more explicit calls to module functions:

```python
import fastapi

router = fastapi.APIRouter()
```

ppr-api/src/main.py
```diff
@@ -2,7 +2,10 @@
 from fastapi import FastAPI, Header, HTTPException
 from starlette.middleware.cors import CORSMiddleware
 
+from endpoints import healthcheck
+
 app = FastAPI()
+app.include_router(healthcheck.router)
```
WalterMoar, line 8: I'm going to change this when I fix my merge conflicts.

ppr-api/src/endpoints/healthcheck.py
jguertin: sure

ppr-api/src/endpoints/healthcheck.py
```diff
@@ -0,0 +1,8 @@
...
+
+
+@router.get("/health", include_in_schema=False)
```
WalterMoar: We should include this endpoint in the OpenAPI docs - we won't expose it through the gateway, but it saves digging in the code when we want to find the check.

When I fix my merge conflicts I'll separate it out in the docs.

WalterMoar (changes requested):
One change with module conventions and readability, to always explicitly use the module name.

ppr-api/src/main.py
jguertin: 👍

ppr-api/src/endpoints/healthcheck.py
jguertin: Sure, if you can separate it out from the docs that's great.  I've removed the flag.

WalterMoar (approved):

WalterMoar:
Gonna merge this so I can grab it.

PR #74, opened 2019-12-13
Add simple search and results page

bryan-gilbert:
The local url is http://localhost:8080/search


bryan-gilbert:
I'm retracting this PR. two reasons. (1) I want to rework the load indicator and (2) the pass through api is nearly ready and I'd like to use it.

PR #76, opened 2019-12-16
Added passthrough search by serial number. #80

WalterMoar:
Added the first endpoint for searching. It's rough but it will work in dev.


bryan-gilbert (approved):
nice and clean

jguertin (approved):
Looks good.  👍

PR #88, opened 2019-12-18
Vue component naming convention

bryan-gilbert:
Fix two files with incorrect names.
Also fix a lint warning


WalterMoar (approved):
Niiiice!

PR #89, opened 2019-12-18
Refactor the PPR API database configuration to leverage dotenv

jguertin:
I refactor the PPR API database configuration to leverage dotenv.

I also removed the constraint that prevented startup of the application if a database password is not provided.  Now, just like any other misconfiguration, it will no longer prevent the app from starting.

@WalterMoar We'll need to re-apply the DC as this gets deployed

### DIFF
--- a/ppr-api/src/config.py
+++ b/ppr-api/src/config.py
@@ -13,3 +13,8 @@ import dotenv
 dotenv.load_dotenv()
 
 IMS_API_URL = os.getenv("PPR_API_IMS_API_URL")
+DB_HOSTNAME = os.getenv('PPR_API_DB_HOSTNAME')
+DB_PORT = int(os.getenv('PPR_API_DB_PORT', '5432'))
+DB_NAME = os.getenv('PPR_API_DB_NAME')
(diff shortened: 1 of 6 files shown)


WalterMoar (approved):

ppr-api/src/config.py
```diff
@@ -13,3 +13,8 @@
 dotenv.load_dotenv()
 
 IMS_API_URL = os.getenv("PPR_API_IMS_API_URL")
+DB_HOSTNAME = os.getenv('PPR_API_DB_HOSTNAME')
+DB_PORT = int(os.getenv('PPR_API_DB_PORT', '5432'))
```
bryan-gilbert, line 17: why provide default for this and not others?

bryan-gilbert (approved):
just one non-blocking question.

ppr-api/src/config.py
WalterMoar: Good point- and why use the plain postgres port rather than the EDB one?

jguertin: Because it does integer parsing and I wanted to avoid an error from a user not providing a value.

PR #90, opened 2019-12-18
Console.log is gone

bryan-gilbert:
Use console.debug for anything deemed worth keeping in to help near term development


WalterMoar (approved):
I still strongly prefer figuring out the debugger. Let's see if we can remove these once that's done.

jguertin (approved):

PR #91, opened 2019-12-18
Make axios call with config

bryan-gilbert:
Instead of hard coding the url query string use the configuration approach

### DIFF
--- a/ppr-ui/src/search/search-serial.ts
+++ b/ppr-ui/src/search/search-serial.ts
@@ -5,7 +5,7 @@ import AppData from '@/utils/app-data'
 function baseUrl(): string {
   const Config = AppData.config
   const ApiUrl = Config.pprApiUrl
-  return ApiUrl + 'search?serial='
+  return ApiUrl + 'search'
 }
 
 const TEXT = {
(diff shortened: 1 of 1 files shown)


jguertin (approved):

jguertin (approved):

ppr-ui/src/search/search-serial.ts
```diff
@@ -5,7 +5,7 @@ import AppData from '@/utils/app-data'
 function baseUrl(): string {
   const Config = AppData.config
   const ApiUrl = Config.pprApiUrl
-  return ApiUrl + 'search?serial='
+  return ApiUrl + 'search'
```
WalterMoar: Any reason to not collapse this into just a return call?

WalterMoar (approved):

ppr-ui/src/search/search-serial.ts
bryan-gilbert: Yes ... can be collapsed.

bryan-gilbert: collapsed in the commit needed to resolve merge conflicts

PR #94, opened 2019-12-18
Removed the OpenShift route. #95

WalterMoar:
Now that the ppr-api is using a service to connect to the ims-api, the route can be removed.

Curious if `oc apply` is enough for this, but it might need to be manually deleted.


jguertin (approved):
Excellent

PR #97, opened 2019-12-19
Added /operations/database endpoint. #96

WalterMoar:
I split the database check out from the regular health check, as we had pod restarts due to db timeouts.

### DIFF
--- a/ppr-api/src/endpoints/healthcheck.py
+++ b/ppr-api/src/endpoints/healthcheck.py
@@ -13,7 +13,7 @@ STATUS_DOWN = "DOWN"
 
 
 @router.get("/health")
-def health(session: sqlalchemy.orm.Session = fastapi.Depends(models.database.get_session)):
+def health():
     """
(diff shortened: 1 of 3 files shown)


ppr-api/src/endpoints/healthcheck.py
```diff
@@ -15,12 +15,20 @@
 @router.get("/health")
 def health(session: sqlalchemy.orm.Session = fastapi.Depends(models.database.get_session)):
```
jguertin: If not using the database in `/health`, you can remove this parameter.

jguertin (approved):
lgtm, just one comment on the health method parameter.  If you like we can remove this later, but would prefer sooner.

ppr-api/src/endpoints/healthcheck.py
WalterMoar: Oopsies.

WalterMoar: Removed.

PR #98, opened 2019-12-19
Database check returns HTTP 500 on fail. #96

WalterMoar:
Changed to database check to return HTTP 500 if it fails.


WalterMoar:
Closing this for now.

PR #190, opened 2019-12-23
Database check returns HTTP 503 on fail. #96

WalterMoar:
Updated the database check to not return 200 on fail. Using a 503 (service unavailable) as a 500 (internal server error) still just doesn't feel right when there is no error internal to the server.

### DIFF
--- a/ppr-api/src/endpoints/healthcheck.py
+++ b/ppr-api/src/endpoints/healthcheck.py
@@ -30,7 +30,7 @@ def database(response: responses.Response,
     Returns a health check for the reachability of the database.
     """
     try:
-        session.execute('SELECT 1')
+        session.execute("SELECT 1")
 
         return {
             "status": STATUS_UP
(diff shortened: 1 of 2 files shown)


bryan-gilbert (approved):
Good to switch to json responses.

ppr-api/src/endpoints/healthcheck.py
```diff
@@ -23,20 +24,23 @@ def health():
...
     Returns a health check for the reachability of the database.
     """
-    return {
-        "status": database_status(session)
-    }
-
-
-def database_status(session: sqlalchemy.orm.Session):
     try:
-        session.execute('SELECT 1')
```
WalterMoar, line 37: For consistency - I'm good with singles or doubles, and prefer doubles, but needs to be consistent.

WalterMoar:
> Good to switch to json responses.

That was Justin - just working from his setup :)

ppr-api/src/endpoints/healthcheck.py
bryan-gilbert: Can the lint tooling fix this? The act of "on save" or "on lint" is the best place to get this consistency.

WalterMoar: Good question, although you may sometimes want to use single quotes if you have a lot of embedded double quotes. PEP8 doesn't pick a lane on this one, it just recommends consistency.

jguertin: I'll screw this up over and over again, lol.  Too much switching between languages.  I'm good with double-quotes, except when they exist in the string.  I'll have to see if I can get the formatter to do this. :)

jguertin (approved):

PR #192, opened 2019-12-24
Add links to home page, update unit test

bryan-gilbert:
Change unit test to use shallowMount to focus on page level components. This is the recommended practice and, besides, with mount and the inclusion of router-link on the page the set up to mount the element becomes very complex.

### DIFF
--- a/ppr-ui/src/views/Home.vue
+++ b/ppr-ui/src/views/Home.vue
@@ -32,13 +32,37 @@
               <header>
                 <h2>Sample Aside Section</h2>
               </header>
+              <div>
+                <ul>
+                  <li>
+                    <router-link to="home">
+                      Home
+                    </router-link>
+                  </li>
+                  <li v-if="flags.feature2">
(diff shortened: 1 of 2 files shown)


ppr-ui/src/views/Home.vue
```diff
@@ -32,13 +32,37 @@
...
                 <h2>Sample Aside Section</h2>
               </header>
+              <div>
+                <ul>
+                  <li>
+                    <router-link to="home">
+                      Home
+                    </router-link>
+                  </li>
+                  <li v-if="flags.feature2">
```
jguertin, line 42: Should we consider removing these feature flags, or at least replacing them with something more descriptive?

jguertin (approved):

ppr-ui/src/views/Home.vue
bryan-gilbert: for example ... ? let's discuss today

PR #193, opened 2019-12-24
Client openshift documentation readme

bryan-gilbert:
Add more details around the process (e.g. logging on, tagging)
Make it easy to copy the command from the readme directly into terminal (e.g. no leading $ prompt indicator and remove redundant folder path)

### DIFF
--- a/ppr-ui/openshift/README.md
+++ b/ppr-ui/openshift/README.md
@@ -10,6 +10,11 @@ oc login https://console.pathfinder.gov.bc.ca:8443 --token=....
 cd ppr-ui/openshift
 ```
 
+## Validation
+
+To any oc apply command you can append ```--validate``` to check for errors in the configuration.  Note that there is a known
+issue with a validation error saying that Route needs a status property. It doesn't so ignore.  
+
 ## Tools Namespace `BuildConfig`s
 
 ```bash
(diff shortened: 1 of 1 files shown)


ppr-ui/openshift/README.md
```diff
@@ -1,28 +1,46 @@
...
-$ oc process -f ppr-ui/openshift/ppr-ui-bc.yaml | oc -n zwmtib-tools apply -f -
-```
+
+## Tagging builds
+
+The following process of tagging is temporary as it will change once we build out our CI/CD pipeline.
+After building an image on the Tools namespace you will want to tag a build with the 'dev', 'test', or ... tag.
+
+```bash
+oc tag zwmtib-tools/ppr-ui:latest ppr-ui:dev
```
jguertin: Would you use `zwmtib-tools/` or `-n zwmtib-tools`?  I'm happy if this works, just didn't know you could do it this way.

jguertin (approved):

ppr-ui/openshift/README.md
bryan-gilbert: The pattern I've seen most often is what is used and not the -n namespace approach. Is the -n approach more common?  I'm ok with consistency and changing this.

PR #194, opened 2019-12-24
Add caddy configuration

bryan-gilbert:
See comments and provided docs for details


ppr-ui/Caddy/Readme.md
```diff
@@ -0,0 +1,14 @@
+# Client UI Caddy
```
jguertin: A subtle file naming convention, but for consistency.  Our other readme's are all uppercase README.md.

ppr-ui/openshift/ppr-ui-bc.yaml
```diff
@@ -28,6 +33,8 @@ objects:
         paths:
         - destinationDir: ./
           sourcePath: /dist
+        - destinationDir: ./
+          sourcePath: /Caddy
```
jguertin, line 37: What does this do, and does it conflict with `dist`?

ppr-ui/openshift/ppr-ui-dc.yaml
```diff
@@ -72,6 +72,10 @@ objects:
             terminationMessagePath: /dev/termination-log
             terminationMessagePolicy: File
             volumeMounts:
+              - mountPath: /etc/Caddyfile
+                name: coops-web-caddy-config
+                readOnly: true
+                subPath: Caddyfile
```
jguertin: Doesn't this essentially replace the Caddyfile you copied in with a different version from the config map?  Also, please consider renaming to remove `coops`.

jguertin (commented):
No specific changes required, but I have a few questions.

ppr-ui/openshift/ppr-ui-bc.yaml
bryan-gilbert: Actually it doesn't. The destinationDir declares what is available for the Docker commands above

ppr-ui/openshift/ppr-ui-dc.yaml
bryan-gilbert: oh oh .. that was a mistake :( . not meant to be left in the code.

ppr-ui/Caddy/Caddyfile
```diff
@@ -0,0 +1,14 @@
...
+log stdout
+
+errors stdout
+
+# Source for the rewrite is from the Vue router docs:
+#https://router.vuejs.org/guide/essentials/history-mode.html#example-server-configurations
+rewrite {
+    regexp .*
+    to {path} /
+}
```
WalterMoar: Can't hurt to put in this newline that GitHub is complaining about.

ppr-ui/Caddy/d-c.yml
```diff
@@ -0,0 +1,10 @@
+version: '3'
+services:
+  caddy:
+    image: "abiosoft/caddy:latest"
```
WalterMoar, line 4: Do we want latest? Ideally we would pull the image from the cluster, but not sure how to do that. Second best would be to match versions of Caddy - although it's a pretty straightforward web server, it more of just a best practice than any kind of concern.

ppr-ui/openshift/ppr-ui-inter-bc.yaml
```diff
@@ -16,6 +16,8 @@ objects:
         kind: ImageStreamTag
         name: ${APP_NAME}:latest
     source:
+      # 1. Checks out the repo; 2 sets the context to be the UI client code; 3 builds the Vue app.
+      # result is an image containing the client code folder and a fully built Vue distribution (in dist folder)
```
WalterMoar, line 20: You're probably aware, but these comments get stripped out by OpenShift. Any export that follows will not have the comments.

WalterMoar:
Looks good :) Some day I'll have to try out the docker stuff.

WalterMoar (approved):

ppr-ui/Caddy/Caddyfile
bryan-gilbert: added

ppr-ui/Caddy/d-c.yml
bryan-gilbert: This docker compose file is for local development so taking latest caddy is sufficient and there is no need to use image: "bcgov-s2i-caddy"

ppr-ui/openshift/ppr-ui-inter-bc.yaml
bryan-gilbert: Right.  Except that we intend to keep our configuration files as the defacto standard and we will not overwrite with an export from OpenShift. We'll cherry pick stuff that we might need.

PR #196, opened 2019-12-24
Serial number search criteria

bryan-gilbert:
Change criteria to allow for up to 25 alpha-numeric characters

### DIFF
--- a/ppr-ui/src/search/search-serial.ts
+++ b/ppr-ui/src/search/search-serial.ts
@@ -7,7 +7,11 @@ function baseUrl(): string {
 }
 
 const TEXT = {
-  errorMsg: (text): string => `Search serial number error - ${text}`
+  errorMsg: (text): string => `Search serial number error - ${text}`,
+  describeValidSerial: 'Serial number can have up to 25 letters or digits',
+  defineValid: 'Serial number must be up to 25 letters or digits',
+  required: 'Require serial number',
+  tooLong: 'Serial number can only have 25 characters'
 }
 
 interface SerialResult {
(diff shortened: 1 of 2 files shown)


jguertin (approved):

ppr-ui/src/search/search-serial.ts
```diff
@@ -7,7 +7,11 @@ function baseUrl(): string {
 }
 
 const TEXT = {
-  errorMsg: (text): string => `Search serial number error - ${text}`
+  errorMsg: (text): string => `Search serial number error - ${text}`,
+  describeValidSerial: 'Serial number can have up to 25 letters or digits',
+  defineValid: 'Serial number must be up to 25 letters or digits',
+  required: 'Require serial number',
+  tooLong: 'Serial number can only have 25 characters'
```
WalterMoar, line 14: I really need to get some time to look at the Composition API and wrap my head around this stuff. I'm not sure what the best practices or conventions are, so I don't know how well I can review this.

However, what I'm thinking is that I'm not sure that the "view", as it were, should be split across the .ts and the .vue files. As I'm reading the template HTML to me it makes sense to just have the text in the template. It simplifies so much. I know that if we do i18n or that kind of thing this is a good framework, but we're probably never going there.  And if we do we can introduce the framework later. Then there's the whole question of the JSON Schema validation, which should be defining these things, but that's a whole other ball of worms.

ppr-ui/src/search/search-serial.ts
```diff
@@ -40,14 +44,15 @@ export class SearcherSerial {
...
-      , (value): (boolean | string) => { return /^\d{4}$/.test(value) || 'serial number must be 4 digits' }
+      (value): (boolean | string) => { return !!value || TEXT.required },
+      (value): (boolean | string) => { return value.length <= 25 || TEXT.tooLong },
+      (value): (boolean | string) => { return /^[0-9a-zA-Z]{1,25}$/.test(value) || TEXT.defineValid }
     ]
   }
 
   public get term(): string { return this._term}
   public get results(): SerialResult[] { return this._results}
+  public get describeValidSerial(): string { return TEXT.describeValidSerial}
```
WalterMoar, line 55: Just my usual concerns about the formatter being used (or not used). Safe to assume this is not VS Code? This will be even more of a problem when Greg comes onboard.

WalterMoar (approved):
Giving the thumbs up on this but I need some time to learn the Composition API and really play with this stuff before I can offer a better review.

ppr-ui/src/search/search-serial.ts
bryan-gilbert: Good point to discuss ...where should text reside when it comes from the logic?  Clearly, text in the html template is UI and is immediately visible to anyone who reads the code.  But text that comes from logic is often buried in the code and mixed in with the code.  This means it is hard to find and maintain by a non-dev team member. For example, a UX or BA who is closer to the end user and wants to know the text presented to the end users is consistent throughout the app.

bryan-gilbert: What we have here is one way to extract the text that is destined to be seen by the user and place it somewhere a UX person can read it.  We might wish to abstract this approach further and centralize all such text in one location.  For now, it may be sufficient to centralize all such text in one location within the logic code.

ppr-ui/src/search/search-serial.ts
bryan-gilbert: When I use VSCode I see no difference.  Let's compare and talk and resolve this.

PR #198, opened 2019-12-31
Test coverage

bryan-gilbert:
Document how to run coverage on the ppr-api and ppr-ui

### DIFF
--- a/ppr-api/README.md
+++ b/ppr-api/README.md
@@ -9,6 +9,98 @@ The PPR API is currently a passthrough API that calls the IMS Transaction Manage
 1. `$ (cd src && uvicorn main:app --reload)`
 1. View the [OpenAPI Docs](http://127.0.0.1:8000/docs).
 
+
+## Developer virtual environment setup
+
+To do development in a virtual environment you need to install the python virtual environment package.
+```bash
+python3 -m venv .venv
+
+# pip may need to be updated ... not sure if it helps to run this outside of the virtual environment, just in case ....
+python3 -m pip install --upgrade pip
+```
+
+### Virtual environment activation and deactivation
+
+Thereafter you can activate the virtual environment with:
+
+```bash
+source ./.venv/bin/activate
+```
+When the virtual environment is active the shell prompt starts ith ```(.venv) ```.  Once inside the virtual environment you can
+run the server, unit tests, code coverage tests and static analysis tools.
+To exit the virtual environment run ```deactivate``` in the virtual environment.
+
+```
+deactivate
+``` 
+
+### Api server
+```
+# In the (.venv) virtual environment ....
+# to run the server
+pip3 install -r requirements/dev.txt
+uvicorn src.main:app --reload
+
+# This should bring the web service up on http://localhost:8000, 
+# with OpenAPI documentation available at http://localhost:8000/docs.
+
+# to exit the app press Ctrl-C
+```
+
+### Static analysis
+
+```
+# In the (.venv) virtual environment ....
+# to run lint and show warnings
+pylint src
+# lint show errors only
+pylint -E src
+# to run the common flake8 static analysis tools 
+flake8 src
+```
+Flake8 (https://pypi.org/project/flake8/) runs
+- PyFlakes (Python error checker)  https://pypi.org/project/pyflakes/
+- pycodestyle (Python style checker) https://pypi.org/project/pycodestyle/
+- Ned Batchelder’s McCabe script (complexity)  https://github.com/PyCQA/mccabe
+
+
+### Unit tests
+```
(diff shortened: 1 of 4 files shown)


rstens:
A few comments:
I think we should document also the generation of the coverage.xml and potentially the coverage html reports.

For python:
The test results should be kept too (only temporarily when generating this for codecov and sonarcloud) by adding --junitxml=testresults.xml to the pyunit statement.
(Should then also .gitignore that file)
Adding **--cov-report html** and/or **--cov-report xml** will get you the files that can then be uploaded to codecov and sonarcloud.

The total line would look like:
`pytest tests --junitxml=testresults.xml --cov=. --cov-report xml tests/unit`

For GUI:
npm run test:unit already seems to generate the clover.xml file. This is needed for CodeCov and sonarcloud as well.

For IMS-API:
Jacoco needs to be added to the build.gradle file.

And then test and coverage are executed with: ./gradlew test jacocoTestReport
````
plugins {
    id 'java'
    id 'application'
    id 'org.springframework.boot' version '2.1.9.RELEASE'
    id 'io.spring.dependency-management' version '1.0.8.RELEASE'
    id 'jacoco'
}

repositories {
    // TODO Artifactory would be more ideal to leverage as a proxy
    jcenter()
}

dependencies {
    implementation 'org.springframework.boot:spring-boot-starter-web'
    implementation 'org.springframework.boot:spring-boot-starter-actuator'
    implementation 'io.jaegertracing:jaeger-client:1.0.0'
    runtimeOnly    'io.opentracing.contrib:opentracing-spring-web-starter:2.1.3'
    implementation 'io.sentry:sentry-spring:1.7.27'
    implementation 'org.apache.commons:commons-lang3:3.9'

    // IBM Libraries required for the implementation. libs extracted from download: https://www.ibm.com/marketing/iwm/iwm/web/preLogin.do?source=imscjd
    // See https://github.com/imsdev/ims-java-springboot/blob/master/ims-springboot-tm/README.md for details
    implementation 'javax.resource:javax.resource-api:1.7.1'
    implementation 'com.ibm.icu:icu4j:59.2'
    // included as local jars for now, but we would be better served to pull the artifact from Artifactory.
    implementation files('libs/ccf2.jar', 'libs/cwybs-adapterfoundation.jar', 'libs/despi.jar', 'libs/imsico.jar', 'libs/marshall.jar')

    testImplementation 'junit:junit:4.12'
    testImplementation 'org.mockito:mockito-core:3.1.0'
    testImplementation 'org.springframework.boot:spring-boot-starter-test'
}

jacocoTestCoverageVerification{}

jacocoTestReport {
    reports {
        xml.enabled true
        html.enabled true
        csv.enabled false
        xml.destination file("${buildDir}/coverage.xml")
        html.destination file("${buildDir}/reports/coverage")
    }
}

application {
    // Define the main class for the application
    mainClassName = 'ca.bc.gov.registries.ppr.App'
}

```

I already discussed this with Justin.

ppr-api/README.md
```diff
@@ -9,6 +9,98 @@ The PPR API is currently a passthrough API that calls the IMS Transaction Manage
...
+
+## Developer virtual environment setup
+
+To do development in a virtual environment you need to install the python virtual environment package.
+```bash
+python3 -m venv .venv
+
+# pip may need to be updated ... not sure if it helps to run this outside of the virtual environment, just in case ....
+python3 -m pip install --upgrade pip
+```
```
WalterMoar, line 21: FYI, there's going to be some overlap with the VS Code install documentation, but that's still in progress.

bryan-gilbert: Sure, would be good to create separate sections.  CLI vs VSCode

ppr-api/README.md
```diff
@@ -9,6 +9,98 @@ The PPR API is currently a passthrough API that calls the IMS Transaction Manage
...
+# .coveragerc to control coverage.py
+[run]
+omit =
+    .venv/*
+```
+To install the package:
+```
+# In the (.venv) virtual environment ....
+pip install pytest-cov
```
WalterMoar, line 95: Should this be part of the dev requirements?

bryan-gilbert: I found that this needed to be run in the v-env prior to running the tests in the virtual environment.

WalterMoar: Yes, but can we include it in requirements/dev.txt so that we install it when we first run `pip3 install -r requirements/dev.txt` ?

WalterMoar: We can just merge this as-is, I'll have a lot of changes to the readme after I complete the VS Code document.

WalterMoar (approved):

ppr-api/README.md
jguertin: Agree, this should be included in `requirements/dev.txt`

PR #199, opened 2020-01-02
Added jacoco reference to build.gradle to do coverage reports for Java…

rstens:
… #168

jacoco coverage test reports can be invoked by `./gradlew jacocoTestReport`

The combination call for test and coverage is : `./gradlew test jacocoTestReport`


jguertin (commented):
👍

jguertin (approved):

PR #202, opened 2020-01-02
Prepare serial model for unit testing

bryan-gilbert:
Removing dependency on imported app data in the model class. Instead require this to be passed into the model class.  This will simplify the structure of unit tests.

### DIFF
--- a/ppr-ui/src/search/search-serial.ts
+++ b/ppr-ui/src/search/search-serial.ts
@@ -1,5 +1,5 @@
 import {inject, provide, reactive} from "@vue/composition-api"
-import axios from "@/utils/axios-auth"
+import axiosAuth from "@/utils/axios-auth"
 import AppData from '@/utils/app-data'
 
 function baseUrl(): string {
@@ -72,7 +72,7 @@ export class SearcherSerial {
     let url = baseUrl + 'search'
     console.log('Make the search api call', url)
     return new Promise((resolve, reject): void => {
-      axios
+      axiosAuth
         .get(url, config)
         .then((response): void => {
           if (response && response.data) {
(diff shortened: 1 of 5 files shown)


ppr-ui/src/search/search-serial.ts
```diff
@@ -54,7 +54,7 @@ export class SearcherSerial {
   public get results(): SerialResult[] { return this._results}
   public get describeValidSerial(): string { return TEXT.describeValidSerial}
 
-  public doSearch(term: string): Promise<string> {
+  public doSearch(baseUrl: string, term: string): Promise<string> {
```
jguertin: Since this is a method on an instance of SearcherSerial, I'd be inclined to make baseUrl a property set on construction, since it shouldn't be different between searches.

jguertin (approved):
One suggested change, looks good.

WalterMoar (approved):

WalterMoar:
Can you connect this PR to the issue?

ppr-ui/src/utils/axios-auth.ts
```diff
@@ -11,15 +11,15 @@ instance.interceptors.request.use(
...
+// instance.interceptors.response.use(
+// // : Promise<AxiosResponse<any>>
+//   (response): AxiosResponse<object> => response,
+//   (error): Promise<void> => Promise.reject(error)
+// )
+//
+//
+// export function setBaseUrl(url: string): void {
+//   axios.defaults.baseURL = url
+// }
```
jguertin: Just curious why you chose to comment rather than remove?

jguertin (approved):
lgtm, just one question

ppr-ui/src/utils/axios-auth.ts
bryan-gilbert: Thanks ... good catch

ppr-ui/src/search/search-serial.ts
bryan-gilbert: Moved to constructor.

PR #213, opened 2020-01-07
Added a port config specific to Patroni.

WalterMoar:
I thought I could re-use the DB_PORT configuration setting, but in OpenShift EDB runs on 5444 but Patroni runs on 5432.


jguertin (approved):
Makes sense

Issue #214, opened 2020-01-07
Search for Financing Statement


jguertin:
As a PPR Web user I would like to submit a search with a Base Registration Number so that I can view a specific existing Financing Statement

As a "searcher" I want to enter the PPR Document Registration Number as a search criterion so I can find any security interests previously registered within the register under that number.

Accentance Criteria:

Acceptance Criteria

GIVEN ... I am a searcher intending to search for a security interest using the number assigned to the record by the PPR when the record was entered into the system
WHEN ... I enter any combination of letters, characters, and numbers representing the PPR Document Registration Number
THEN ...
I can get my search criterion validated against applicable rules, with an option to proceed regardless of the results of the validation
I am charged an appropriate amount for this search, with a visible indication if this search will cost me money
I can choose to cancel the search at any point
I have the ability to pay for this search using a one-time credit card payment so I can get my search results
I get a list of exact matches (if a security interest exists) for review, and no matches otherwise
I can select (or have the system automatically select) any matches
I can review my selected matches/search results
I can export my search results to PDF
I can print my search results
I can email my search results
Dependencies? (If so, link dependency in the ticket, make it visible in a team´s backlog)

Validation Rules? (If yes, list here)

A valid PPR Document Registration Number is a combination of letters, characters, and numbers
A valid PPR Document Registration Number is exactly 7 characters long

WalterMoar:
Can we call this done?

jguertin:
The actual search feature of searching by Registration Number has been completed.  Some general search rules (what constitutes being available for search, for examples) may be outstanding, but those will be covered with a broader story, #726.

There were some AC in here, that didn't make much sense within the context of the goal when registration number search was done, and should be handled as independent stories:
- I get a list of exact matches (if a security interest exists) for review, and no matches otherwise
- I can select (or have the system automatically select) any matches
- I can review my selected matches/search results
- I can export my search results to PDF
- I can print my search results
- I can email my search results

PR #215, opened 2020-01-07
Refine the search API interface to re-use a single field and vary by type

jguertin:
Refine the search API interface to re-use a single field and vary by type, rather than an individual criteria for each search type.

The reasoning is that except for debtor name searches, all search types only allow a single field value, so it felt overly complicated to provide a single criteria field for each search type.  Rather, the client can specify the type by enumeration and provide the field value in the available `criteria.value` field.  For individual debtor, the client will still be required to break it down as last name and first name are both required and independent.

This change is to inform the implementation of #214.

Recommended tool for reviewing this file is [Stoplight Studio](https://stoplight.io/)


WalterMoar (approved):

bryan-gilbert (approved):
as discussed .. good idea

PR #216, opened 2020-01-07
Introduce alembic to PPR API

jguertin:
Introduce alembic to PPR API in order to support addition of search table

New files in this commit were generated by running `alembic init migrations`

### DIFF
--- a/ppr-api/migrations/README
+++ b/ppr-api/migrations/README
@@ -1,0 +1,1 @@
+Generic single-database configuration.
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 __pycache__
 .pytest_cache
 .eggs
+*.egg-info/
 
 # Environments
 .venv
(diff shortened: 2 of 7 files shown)


WalterMoar (approved):
Great? I'm not familiar with Alembic so it's all magic to me.

ppr-api/migrations/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
WalterMoar, line 1: GitHub seems really picky about wanting a newline on the last line.

bryan-gilbert:
> Great? I'm not familiar with Alembic so it's all magic to me.

ditto for me

ppr-api/migrations/README
jguertin: Will correct with the next PR.  I'm going to add some alembic usage notes with the first migration.

PR #254, opened 2020-01-10
Sonar runner and accompanying jenkins files for running in openshift

rstens:
For future consideration. The image that runs this needs to be updated. I left a request in the appropriate project. My work is done for now.


bryan-gilbert (approved):
I'm not familiar with jenkins so can't comment on the details of this PR.  That said if I had to maintain this project I wonder how much of the code is boilerplate from some source that I could use to check for updates and how much is code that is owned by the project.

jguertin (commented):
@rstens I've had a look through, but I'm not clear on the value of committing this work as we are not using Jenkins at this point.

rstens:
> @rstens I've had a look through, but I'm not clear on the value of committing this work as we are not using Jenkins at this point.

I am starting to use Jenkins in the tool area as this is my main enabler for running the test tools, sonarqube etc. Tools has Jenkins installed (as an artifact of creating the first pipeline there.)

This is part of setting up the infrastructure so that I can do my work.

PR #260, opened 2020-01-14
Revise the initial search table migration

jguertin:
After some review, made some changes to the search table definition:
- Remove the individual criteria columns in favour of a JSON column for greater flexibility
- Create a code table to hold search types

Normally, I would not update a migration after it has merged, but since this one hasn't been deployed anywhere yet, it was simpler to make the change in this case.

### DIFF
--- a/ppr-api/migrations/versions/c17a9c638a89_create_search_table.py
+++ b/ppr-api/migrations/versions/c17a9c638a89_create_search_table.py
@@ -18,17 +18,33 @@ depends_on = None
 
 
 def upgrade():
+    search_type_table = op.create_table(
+        'search_type',
+        sa.Column('short_code', sa.CHAR(length=2), primary_key=True),
+        sa.Column('long_code', sa.String(length=40), nullable=False),
+        sa.Column('description', postgresql.TEXT, nullable=False),
+    )
+
     op.create_table(
         'search',
-        sa.Column('id', postgresql.UUID, primary_key=True),
-        sa.Column('type', sa.String(length=40), nullable=False),
-        sa.Column('criteria_value', postgresql.TEXT),
-        sa.Column('debtor_first_name', postgresql.TEXT),
-        sa.Column('debtor_second_name', postgresql.TEXT),
-        sa.Column('debtor_last_name', postgresql.TEXT),
-        sa.Column('creation_date_time', sa.DateTime(timezone=True), server_default=sa.text("NOW()"), nullable=False)
+        sa.Column('id', sa.BigInteger, primary_key=True),
+        sa.Column('type_short_code', sa.CHAR(length=2), sa.ForeignKey('search_type.short_code'), nullable=False),
+        sa.Column('criteria', postgresql.JSON, nullable=False),
+        sa.Column('creation_date_time', sa.DateTime(timezone=True), server_default=sa.text('NOW()'), nullable=False)
     )
+
+    # Populate the search type code tables.  Short codes are reflective of the codes in the original PPR system,
+    # while the long codes are for use in the API moving forward.
+    op.bulk_insert(search_type_table, [
+        {'short_code': 'AS', 'long_code': 'AIRCRAFT_DOT', 'description': 'Aircraft Airframe D.O.T. Number'},
+        {'short_code': 'BS', 'long_code': 'BUSINESS_DEBTOR', 'description': 'Business Debtor Name'},
+        {'short_code': 'IS', 'long_code': 'INDIVIDUAL_DEBTOR', 'description': 'Individual Debtor Name'},
+        {'short_code': 'MS', 'long_code': 'MHR_NUMBER', 'description': 'Manufactured Home Registration Number'},
+        {'short_code': 'RS', 'long_code': 'REGISTRATION_NUMBER', 'description': 'Registration Number'},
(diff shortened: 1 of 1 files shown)


ppr-api/migrations/versions/c17a9c638a89_create_search_table.py
```diff
@@ -18,17 +18,33 @@
...
-        sa.Column('type', sa.String(length=40), nullable=False),
-        sa.Column('criteria_value', postgresql.TEXT),
-        sa.Column('debtor_first_name', postgresql.TEXT),
-        sa.Column('debtor_second_name', postgresql.TEXT),
-        sa.Column('debtor_last_name', postgresql.TEXT),
-        sa.Column('creation_date_time', sa.DateTime(timezone=True), server_default=sa.text("NOW()"), nullable=False)
+        sa.Column('id', sa.BigInteger, primary_key=True),
+        sa.Column('type_short_code', sa.CHAR(length=2), sa.ForeignKey('search_type.short_code'), nullable=False),
+        sa.Column('criteria', postgresql.JSON, nullable=False),
+        sa.Column('creation_date_time', sa.DateTime(timezone=True), server_default=sa.text('NOW()'), nullable=False)
```
WalterMoar, line 33: Does `timezone=True` store dates as Pacific Time? The cluster runs as UTC, and we store dates as UTC, and then do all the TZ adjustments for display in PT, and such.

ppr-api/migrations/versions/c17a9c638a89_create_search_table.py
```diff
@@ -18,17 +18,33 @@
...
     )
 
+    # Populate the search type code tables.  Short codes are reflective of the codes in the original PPR system,
+    # while the long codes are for use in the API moving forward.
```
WalterMoar, line 37: Nice.

ppr-api/migrations/versions/c17a9c638a89_create_search_table.py
```diff
@@ -18,17 +18,33 @@
 
 
 def upgrade():
+    search_type_table = op.create_table(
+        'search_type',
+        sa.Column('long_code', sa.String(length=40), primary_key=True),
+        sa.Column('short_code', sa.CHAR(length=2), nullable=False),
```
bryan-gilbert, line 24: Curious, why is the long the foreign key and not the short.  Are the short codes not unique?

ppr-api/migrations/versions/c17a9c638a89_create_search_table.py
bryan-gilbert: Ah that explains why the long codes. Yes. Nice :)

ppr-api/migrations/versions/c17a9c638a89_create_search_table.py
bryan-gilbert: so below

ppr-api/migrations/versions/c17a9c638a89_create_search_table.py
```diff
@@ -18,17 +18,33 @@
...
 
+    # Populate the search type code tables.  Short codes are reflective of the codes in the original PPR system,
+    # while the long codes are for use in the API moving forward.
+    op.bulk_insert(search_type_table, [
+        {'long_code': 'AIRCRAFT_DOT', 'short_code': 'AS', 'description': 'Aircraft Airframe D.O.T. Number'},
+        {'long_code': 'BUSINESS_DEBTOR', 'short_code': 'BS', 'description': 'Business Debtor Name'},
+        {'long_code': 'INDIVIDUAL_DEBTOR', 'short_code': 'IS', 'description': 'Individual Debtor Name'},
+        {'long_code': 'MHR_NUMBER', 'short_code': 'MS', 'description': 'Manufactured Home Registration Number'},
+        {'long_code': 'REGISTRATION_NUMBER', 'short_code': 'RS', 'description': 'Registration Number'},
+        {'long_code': 'SERIAL', 'short_code': 'SS', 'description': 'Serial Number'}
```
bryan-gilbert: Could we use "SERIAL_NUMBER instead of just SERIAL to make searching for uses of the short code easier?

bryan-gilbert (approved):

ppr-api/migrations/versions/c17a9c638a89_create_search_table.py
jguertin: Honestly, I'm not 100% clear on the answer.  I used True in this case to be consistent with lear.  I intend to explore a bit deeper when we do the POST to confirm what timezone we get on the record.

The field defaults to `NOW()` in the database, which I'm pretty sure resolves to UTC.  TZ adjustment into PST should be a client side action.

You raise a good point about dates, and we should be very careful to not use `datetime` fields for actual dates (expiry dates, birthdays, etc...) as these will cause us pain.

WalterMoar: It could be that it just means store the timezone, which would be UTC. We'll see timezone issues right away.

jguertin: As discussed, I'll leave as is for now.  This post gives a pretty decent explanation of the difference. 
 https://stackoverflow.com/questions/5876218/difference-between-timestamps-with-without-time-zone-in-postgresql has a good run down

Issue #242, opened 2020-01-09
ppr-api: Get a search record


jguertin:
As an API User, given the identifier for a search record, retrieve the details of the search.  The response should include:
- The type of search 
- Criteria
- When the search record was created
- A reference to the search results resource 

Implement a new endpoint at `GET /searches/{searchId}`

Refer to the [OAS API Contract](https://github.com/bcgov/ppr/blob/master/ppr-api/docs/api_contract/ppr-api.v1.yaml) for specifics.  This is subject to change.

PR #262, opened 2020-01-14
Get single search record

jguertin:
Added an endpoint to pull a search record out of the db: `GET /searches/{id}`

I'm going to work next on adding some unit tests.  To support this, expect some refactoring for dependency injection as per https://fastapi.tiangolo.com/tutorial/dependencies/first-steps/

If you wish, you can create some sample data by running inserts similar to this after running the `alembic` scripts:
```
insert into search (type_long_code, criteria) values ('INDIVIDUAL_DEBTOR', '{"debtorName":{"first":"Homer","second":"J","last":"Simpson"}}');
insert into search (type_long_code, criteria) values ('REGISTRATION_NUMBER', '{"value": "123456A"}');
```

### DIFF
--- a/ppr-api/src/endpoints/search.py
+++ b/ppr-api/src/endpoints/search.py
@@ -1,14 +1,15 @@
 """ Define the endpoints for searching. """
 
 import json
-from uuid import UUID
 
 import fastapi
 import requests
+import sqlalchemy.orm
 from starlette import responses
 
 import config
 import schemas.search
+import models.search
 
 router = fastapi.APIRouter()
 
@@ -29,8 +30,8 @@ async def search(serial: str, response: responses.Response):
     return json.loads(ims_response.content.decode("utf-8"))
 
 
-@router.get("/searches/{search_id}", response_model=schemas.search.Search)
-def read_search(search_id: UUID):
+@router.get("/searches/{search_id}", response_model=schemas.search.Search, response_model_by_alias=False)
+def read_search(search_id: int, session: sqlalchemy.orm.Session = fastapi.Depends(models.database.get_session)):
     """
     Get the details for a previously submitted search request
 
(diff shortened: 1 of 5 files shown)


ppr-api/src/endpoints/search.py
```diff
@@ -25,3 +28,19 @@
     response.status_code = ims_response.status_code
 
     return json.loads(ims_response.content.decode("utf-8"))
+
+
+@router.get("/searches/{search_id}", response_model=schemas.search.Search, response_model_by_alias=False)
+def read_search(search_id: int, session: sqlalchemy.orm.Session = fastapi.Depends(models.database.get_session)):
```
WalterMoar, line 34: Just want to make sure you're OK with using EDB for this. In addition to instability, it means we won't be able to deploy beyond dev.

jguertin: Good point.  How long do you imagine the patroni instance hanging around?  Should we just switch the database environment over to patroni?

Personally, I don't have a strong feeling, except I'd rather not retroactively go and change all our services from one module to another when we finally settle on one...

What if we consolidate the code to a single database configuration but create a second deployment in dev that points to EDB with ppr-api-dev sticking to Patroni?

WalterMoar: Many ways to slice a cat! (Or whatever the saying is).

My concern about two APIs writing to two DBs is that people will be somewhat confused. Plus then we have two databases that we are trying to keep up to date with changes.

I wonder if we should:
 - copy the existing models.database to models.edb.
 - retain the existing models.patroni.
 - change models.database to be a thin layer over models.patroni for now. That way all the imports can be for models.database. Ignore EDB until it is up and running, then we can turf models.patroni and copy models.edb to models.database (and do an export/import from Patroni into EDB and decommission Patroni).

Thoughts?

WalterMoar: (keeping in mind that the only difference between the EDB and Patroni connections is the service and port being called).

jguertin: Yeah, that sounds workable.  I can take a run at that.

WalterMoar:
The dependency list says this issue is blocked by #242. Is the dependency backwards? (should be blocking, not blocked)

jguertin:
> The dependency list says this issue is blocked by #242. Is the dependency backwards? (should be blocking, not blocked)

Yeah, I accidentally linked it as a dependency instead of the connect button, and I don't appear to have the ability to remove the dependency. 😡

jguertin:
> The dependency list says this issue is blocked by #242. Is the dependency backwards? (should be blocking, not blocked)

I finally found the obscure remove link and took out the dependency

WalterMoar (approved):
Looks good - the EDB/Patroni is good going in here, or another PR, or whatever.

PR #263, opened 2020-01-15
Make the default database module access Patroni instead of EDB

jguertin:
@WalterMoar A few points on this one:
- the `/operations/database` endpoint will now use patroni, so you'll need to use `/operations/edb` for the EDB database
- I set `DB_HOSTNAME_PATRONI` to default to `DB_HOSTNAME` if the env variable is not set, this is so working locally will work ongoing, but still leverages the same config available in our environments.

### DIFF
--- a/ppr-api/src/config.py
+++ b/ppr-api/src/config.py
@@ -13,7 +13,7 @@ import dotenv
 dotenv.load_dotenv()
 
 DB_HOSTNAME = os.getenv('PPR_API_DB_HOSTNAME')
-DB_HOSTNAME_PATRONI = os.getenv('PPR_API_DB_HOSTNAME_PATRONI')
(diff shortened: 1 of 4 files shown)


ppr-api/src/config.py
```diff
@@ -13,7 +13,7 @@
 dotenv.load_dotenv()
 
 DB_HOSTNAME = os.getenv('PPR_API_DB_HOSTNAME')
-DB_HOSTNAME_PATRONI = os.getenv('PPR_API_DB_HOSTNAME_PATRONI')
+DB_HOSTNAME_PATRONI = os.getenv('PPR_API_DB_HOSTNAME_PATRONI', DB_HOSTNAME)
```
WalterMoar, line 16: Nice!

WalterMoar (dismissed):

WalterMoar (approved):

PR #264, opened 2020-01-16
Refactor search endpoint to support dependency injection

jguertin:
Used FastAPIs dependency injection mechanism for accessing the database.  Among other benefits this provides improved support for mocking in unit tests.  An overview on the topic can be found here: ["Dependency Injection"?](https://fastapi.tiangolo.com/tutorial/dependencies/first-steps/#dependency-injection)

This is a departure from the way things are done in `sbc-auth` and `lear`, as both seem to rely heavily on static methods, which are not terribly conducive to this approach.

### DIFF
--- a/ppr-api/src/endpoints/search.py
+++ b/ppr-api/src/endpoints/search.py
@@ -4,12 +4,11 @@ import json
 
 import fastapi
 import requests
-import sqlalchemy.orm
 from starlette import responses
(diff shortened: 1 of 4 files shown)


WalterMoar (approved):

bryan-gilbert (approved):
I like this

PR #265, opened 2020-01-16
Add integration tests to the PPR API

jguertin:
Create the first integration test for the PPR API.  This will allow us to test the API interface through to a database directly.

A couple of points to note about this PR:
- The test creates its own data and does not clean it up.  If this concerns you, lets chat.
- I used the "Compile and Unit Test" workflow for the time being, since the integration tests are still pretty quick.  The job took 1:08 to complete on my fork.  We might as well keep it here for now and isolate them out to a different workflow if things slow down.
- @rstens For the moment, these tests are running after code coverage is pushed up, we can chat a bit about merging the results later
- FYI, I had initially included running the alembic scripts in `ready_local_db.sh`, but then I had to make assumptions about whether or not the script was being run in a virtual environment, so I took it out and ran alembic in isolation.

Enjoy.

### DIFF
--- a/.github/workflows/compile-and-test-stage.yml
+++ b/.github/workflows/compile-and-test-stage.yml
@@ -6,6 +6,14 @@ jobs:
   ppr-api-build:
     runs-on: ubuntu-latest
 
+    env:
+      PPR_API_DB_HOSTNAME: localhost
+      PPR_API_DB_PORT: 5432
+      PPR_API_DB_NAME: ppr
+      PPR_API_DB_USERNAME: postgres
+      PPR_API_DB_PASSWORD: int_test_pw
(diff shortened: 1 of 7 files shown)


rstens (commented):
We could potentially do the unit tests and the integration tests in **one go** and then run codecov.
Or keep them separate and run run codecov after the integrations tests again as we need to because the coverage files will be overwritten every single time.

gh2os (approved):

bryan-gilbert (approved):
nice

WalterMoar (approved):
Looks good - didn't get a chance to play with it, but it's been sitting there since last week so let's ship it.

PR #268, opened 2020-01-17
Fix resources in the intermediate build config

bryan-gilbert:


ppr-ui/openshift/ppr-ui-bc.yaml
```diff
@@ -40,6 +40,7 @@ objects:
       dockerStrategy:
         from:
           kind: ImageStreamTag
+          # TODO: research if we should specify version rather than take latest ....
```
jguertin: Do you have a ticket for this? If so, it's a good idea to include the ticket number in the TODO.

bryan-gilbert: meant to do this ... thanks for catching it ... see recent commit

ppr-ui/openshift/ppr-ui-bc.yaml
```diff
@@ -40,6 +40,7 @@ objects:
       dockerStrategy:
         from:
           kind: ImageStreamTag
+          # TODO: Issue 273 research if we should specify version rather than take latest ....
```
WalterMoar, line 43: Awesome - if you can add a sticky to the techdebt "radiator" that would be awesome too.

WalterMoar (approved):

PR #269, opened 2020-01-17
Ignore all files prefixed with .env

bryan-gilbert:
let's just make a blanket rule.  This is helpful for me with another PR which uses three .env.* files.


rstens (approved):
Makes sense.

jguertin (approved):

gh2os (approved):

PR #271, opened 2020-01-17
Deploy PPR UI on bcregistry.ca

bryan-gilbert:
To get here we need to make several adjustments in many different locations. This commit strives to reduce the number of moving parts to one environment variable that is pushed into the build configurations and then into the running environment. This variable is used by vue.config.js which provides it to the application BASE_URL which shares it with the rest of the application.

As well this commit provides a standard way to process and apply the build and deploy configs.  This standardizes our default configuration (just run the process.sh script) and allows for experimentation (e.g. load from a private repo branch and create new openshift resources that don't touch our running app) and for a rapid change to a new domain or subpath (we expect to change from /cooperatives/ppr to /ppr sometime in the near term).

### DIFF
--- a/ppr-ui/README.md
+++ b/ppr-ui/README.md
@@ -30,7 +30,11 @@ npm run lint
 ```
 
 ### Customize configuration
-See [Configuration Reference](https://cli.vuejs.org/config/).
+The PPR web application is currently served via https://bcregistry.ca/cooperatives/ppr. This url represents
+some technical debt and will need to be changed. The subpath ```cooperatives/ppr```
+appears in the three Openshift templates as APP_PATH; also in ```vue.config.js``` and in ```src/utils/config-helper.ts```
+
+Also see [CLI Vue Configuration Reference](https://cli.vuejs.org/config/).
 
 ## Debugging
 
--- a/ppr-ui/openshift/README.md
+++ b/ppr-ui/openshift/README.md
@@ -10,6 +10,8 @@ oc login https://console.pathfinder.gov.bc.ca:8443 --token=....
 cd ppr-ui/openshift
 ```
 
+Take a close look at ```process.sh`` to see how to set up parameters for build and deploy configs.
+
 ## Validation
 
 To any oc apply command you can append ```--validate``` to check for errors in the configuration.  Note that there is a known
--- a/ppr-ui/openshift/ppr-ui-dc.yaml
+++ b/ppr-ui/openshift/ppr-ui-dc.yaml
@@ -7,6 +7,7 @@ objects:
     labels:
       app: ppr
       deploymentconfig: ${APP_NAME}
+      pprgroup: ${PPR_GROUP}
       role: ${APP_NAME}
     name: ${APP_NAME}
   spec:
@@ -15,6 +16,7 @@ objects:
     selector:
       app: ppr
       deploymentconfig: ${APP_NAME}
+      pprgroup: ${PPR_GROUP}
       environment: ${ENVIRONMENT}
       role: ${APP_NAME}
     strategy:
@@ -32,6 +34,7 @@ objects:
         labels:
           app: ppr
           deploymentconfig: ${APP_NAME}
+          pprgroup: ${PPR_GROUP}
           environment: ${ENVIRONMENT}
           role: ${APP_NAME}
       spec:
@@ -42,7 +45,7 @@ objects:
             livenessProbe:
               failureThreshold: 3
               httpGet:
-                path: /
+                path: ${APP_PATH}
                 port: 2015
                 scheme: HTTP
               periodSeconds: 10
@@ -55,7 +58,7 @@ objects:
             readinessProbe:
               failureThreshold: 3
               httpGet:
-                path: /
+                path: ${APP_PATH}
                 port: 2015
                 scheme: HTTP
               initialDelaySeconds: 5
@@ -72,9 +75,13 @@ objects:
             terminationMessagePath: /dev/termination-log
             terminationMessagePolicy: File
             volumeMounts:
-              - mountPath: /var/www/html/config
-                name: ppr-web-ui-config
+              - mountPath: /var/www/html${APP_PATH}/config
+                name: ${APP_CONFIG}
                 readOnly: true
+              - mountPath: /etc/Caddyfile
+                name: ${CADDY_CONFIG}
+                readOnly: true
+                subPath: Caddyfile
         dnsPolicy: ClusterFirst
         restartPolicy: Always
         schedulerName: default-scheduler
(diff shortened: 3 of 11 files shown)


ppr-ui/openshift/ppr-ui-dc.yaml
```diff
@@ -135,7 +149,30 @@ objects:
       name: ${APP_NAME}
       weight: 100
     wildcardPolicy: None
+- apiVersion: v1
+  kind: ConfigMap
```
WalterMoar, line 153: Nice to see the configmap in the dc (although at some point we have to admit this file is the "deploy" not the "dc").

ppr-ui/openshift/process.sh
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env bash
+
+# Git repo values as used in the yaml config files:
+#REPO=https://github.com/bcgov/ppr
+#BRANCH=master
```
jguertin: Should probably set this back to bcgov/ppr `master` before merging.

ppr-ui/openshift/process.sh
```diff
@@ -0,0 +1,97 @@
...
+APP_PATH=$APP_PATH
+CADDY_CONFIG=$APP_NAME-caddy-config$APP_TAG
+IMAGE_TAG=latest
+ROUTE_HOST=$HOST
+" > .env.dc
+
+# Process and apply the build configs
+echo "oc process -f ppr-ui-inter-bc.yaml --param-file=.env.inter | oc -n zwmtib-tools apply -f -"
+echo ".env.inter"
+cat .env.inter
```
jguertin, line 69: Is there any sensitive information in this file?  If so, probably shouldn't `cat` to the console.

ppr-ui/README.md
```diff
@@ -30,7 +30,11 @@ npm run lint
 ```
 
 ### Customize configuration
-See [Configuration Reference](https://cli.vuejs.org/config/).
+The PPR web application is currently served via https://bcregistry.ca/cooperatives/ppr. This url represents
+some technical debt and will need to be changed. The subpath ```cooperatives/ppr```
+appears in the three Openshift templates as APP_PATH; also in ```vue.config.js``` and in ```src/utils/config-helper.ts```
+
+Also see [CLI Vue Configuration Reference](https://cli.vuejs.org/config/).
```
WalterMoar: As discussed, I don't think that this is the approach we want to take. A custom script for this is one more thing to learn, troubleshoot, and maintain. Understood if you find it useful, but I don't think we want it as part of the repo.

ppr-ui/openshift/README.md
```diff
@@ -10,6 +10,8 @@ oc login https://console.pathfinder.gov.bc.ca:8443 --token=....
 cd ppr-ui/openshift
 ```
 
+Take a close look at ```process.sh`` to see how to set up parameters for build and deploy configs.
```
WalterMoar, line 13: As discussed, I don't think that this is the approach we want to take. A custom script for this is one more thing to learn, troubleshoot, and maintain. Understood if you find it useful, but I don't think we want it as part of the repo.

ppr-ui/vue.config.js
```diff
@@ -1,3 +1,9 @@
+
+// For local development set the APP_PATH environment variable to '/'
+// or use  'npm run serve'  (package.json)
+const APP_PATH = process.env.APP_PATH || '/cooperatives/ppr'
+console.log('Start Vue build (vue.config.js) with APP_PATH', APP_PATH)
```
jguertin: prefer `console.info` or `console.debug` over `console.log`

ppr-ui/openshift/ppr-ui-dc.yaml
```diff
@@ -15,6 +16,7 @@ objects:
     selector:
       app: ppr
       deploymentconfig: ${APP_NAME}
+      pprgroup: ${PPR_GROUP}
       environment: ${ENVIRONMENT}
```
WalterMoar, line 20: I don't think we need another label & selector - can the environment selector not be used to accomplish what is needed?

jguertin (changes requested):
Looks good, just looking for a bit of cleanup, primarily with regards to using the master based settings rather than referring to your own temporary settings.

ppr-ui/src/utils/config-helper.ts
```diff
@@ -1,42 +1,24 @@
...
-    // remove next line once above to do is done
-    const url = '/config/configuration.json'
-
+    const url = `${APP_PATH}config/configuration.json`
     const headers = {
       'Accept': 'application/json',
       'ResponseType': 'application/json',
       'Cache-Control': 'no-cache'
     }
+    // console.log('Retrieve configuration from url:', url)
```
WalterMoar: We shouldn't be committing commented-out code.

ppr-ui/openshift/ppr-ui-dc.yaml
bryan-gilbert: No. the environment does not uniquely identify all the elements added to both the build and deploy related to the deploy.  The purpose of this label is to safely call oc delete and remove all of those elements.   Then purpose of the environment variable is to tell the system what "environment" (dev, test, prod) is active.

ppr-ui/openshift/process.sh
bryan-gilbert: Our practice and implementation to date has been to not include sensitive information in these files.  I expect we will continue to stick with this practice.

ppr-ui/openshift/README.md
bryan-gilbert: As discussed, this is a step in a progression of steps.  We're moving towards building on github and when we do these concerns over build scripts will go away.  Yet the new approach will need to allow for the ability to create experimental/pr based builds.  Until then a script like this is better than a ton of documentation. 
Note that the use of the script is not required for "our main" deployments because our main configurations are the default values in the yaml files. The script is for experimentation or pr based adjustments.  Although the script is not required, it does make it easier to redeploy "our main" deployments because it encapsulates four tasks into one.

ppr-ui/src/utils/config-helper.ts
bryan-gilbert: Removed

ppr-ui/openshift/process.sh
bryan-gilbert: done

ppr-ui/vue.config.js
bryan-gilbert: done

jguertin (approved):

bryan-gilbert:
Created task #281 to track a discussion around use of openshift label to unify all related resources

Issue #20, opened 2019-11-01
ppr-api CORS configuration


WalterMoar:
## Background

We need to determine how to set up the Cross-Origin Resource Sharing (CORS - https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) configuration for the API.

## Details

We need to make this configurable at runtime, probably using the standard `dotenv` and `os.getenv` that we see in other BCRS apis. The configuration is currently hard-coded but we should probably go with:
 - dev: http://localhost:8080 (desktop frontend) and https://ppr-dev.pathfinder.gov.bc.ca (dev frontend).
 - test: https://ppr-test.pathfinder.gov.bc.ca (test frontend).
 - prod: https://ppr.pathfinder.gov.bc.ca (prod frontend).

This will probably all change with apigee (or other gateway) but for now it is what we know.

WalterMoar:
This is still important, but we aren't going to be working on the Python API for the next chunk of work. Closing this.

WalterMoar:
Re-opening this now that we're focusing on the Python API.

bryan-gilbert:
Need to account for changing ports on localhost.  May be less of an issue once the IMS API is shutdown. Still may wish to include, by default in our code, the following for localhost
    "http://localhost:8080",
    "http://localhost:8081",
    "http://localhost:8000",
    "http://localhost:8001",

PR #275, opened 2020-01-20
Update main.py

tendayichakanyuka:
The reason this code change is needed is to allow our PPR API to accept incoming requests from our new domain on bcregistry.ca.

### DIFF
--- a/ppr-api/src/main.py
+++ b/ppr-api/src/main.py
@@ -13,7 +13,7 @@ app.include_router(api.router)
 
 origins = [
     "http://localhost:8080",
-    "https://ppr-dev.pathfinder.gov.bc.ca"
+    "https://ppr-dev.pathfinder.gov.bc.ca",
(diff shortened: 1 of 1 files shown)


ppr-api/src/main.py
```diff
@@ -14,6 +14,7 @@
 origins = [
     "http://localhost:8080",
     "https://ppr-dev.pathfinder.gov.bc.ca"
+    "https://dev.bcregistry.ca"
```
WalterMoar: There's a missing comma on line 16, which is going to cause these two strings to be concatenated.

WalterMoar:
Note that we also have #20, which will either turf these settings completely, or move them into an environment variable so that we don't need a code change for this kind of thing. (tech debt)

jguertin (approved):

WalterMoar (approved):
PR from the PO is go 😄

PR #278, opened 2020-01-20
Convert search.criteria from JSON to JSONB

jguertin:
Upon advice from Bob, now using JSONB over JSON in postgres.  The differences are highlighted at https://www.postgresql.org/docs/11/datatype-json.html


WalterMoar (approved):

bryan-gilbert (approved):
For clarity, lets note that the storage will now be a JSON object as opposed to JSON text.
Good choice.

PR #279, opened 2020-01-20
UI - Update node packages - Issue 256

bryan-gilbert:
performed manual and auto fixes
npm audit fix # fixed one of 7
npm install --save @vue/cli  # fixed second of 7
npm audit fix # fixed remaining 5

### DIFF
--- a/ppr-ui/package.json
+++ b/ppr-ui/package.json
@@ -26,7 +26,6 @@
   "dependencies": {
     "@sentry/browser": "^5.9.1",
     "@sentry/integrations": "^5.8.0",
-    "@vue/cli": "^4.1.2",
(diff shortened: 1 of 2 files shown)


ppr-ui/package.json
```diff
@@ -26,6 +26,7 @@
   "dependencies": {
     "@sentry/browser": "^5.9.1",
     "@sentry/integrations": "^5.8.0",
+    "@vue/cli": "^4.1.2",
```
jguertin: Shouldn't `@vue/cli` be in `devDependencies`? Don't think it should be used in production code.

bryan-gilbert: Yup.  Moved

jguertin (approved):

PR #280, opened 2020-01-20
Post search endpoint

jguertin:
Created an end-point to handle a POST requests to `/searches`.  This included a few changes including:
- The new "POST" method in the end point
- Addition of commit/rollback around the database dependency
- Implementation of pydantic validators in the schema. Ran into one odd-ball issue here:
  - The validator should act like a class method and doesn't require self.  pylint worked fine on my machine, but failed in Github Actions: https://github.com/vituary-solutions/ppr/runs/399781757
  - To work around this for now, I've added hints for pylint to ignore it
- Updates to the model and repository for creating database objects
- unit and integration tests

### DIFF
--- a/ppr-api/src/endpoints/search.py
+++ b/ppr-api/src/endpoints/search.py
@@ -4,7 +4,7 @@ import json
 
 import fastapi
 import requests
-from starlette import responses
+from starlette import responses, status
 
 import config
 import schemas.search
@@ -43,3 +43,12 @@ def read_search(search_id: int, search_repository: repository.search_repository.
     if search_db_row is None:
         raise fastapi.HTTPException(status_code=404, detail="Search record not found")
     return search_db_row
+
+
+@router.post("/searches", response_model=schemas.search.Search, response_model_by_alias=False)
+def create_search(response: responses.Response, search_input: schemas.search.SearchBase,
+                  search_repository: repository.search_repository.SearchRepository = fastapi.Depends()):
+    # TODO execute the search query and create search_result records. See https://github.com/bcgov/ppr/issues/222
+    search_model = search_repository.create_search(search_input)
+    response.status_code = status.HTTP_201_CREATED
+    return search_model
--- a/ppr-api/src/endpoints/search.py
+++ b/ppr-api/src/endpoints/search.py
@@ -14,7 +14,7 @@ router = fastapi.APIRouter()
 
 
 # TODO: read the ims-api endpoint from an environment variable.
-@router.get("/search")
+@router.get('/search')
 async def search(serial: str, response: responses.Response):
     """
     Find financial statements that match the search criteria.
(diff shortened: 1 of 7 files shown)


ppr-api/src/schemas/search.py
```diff
@@ -1,14 +1,47 @@
...
+        except KeyError:
+            raise ValueError('type must be one of: {}'.format(list(map(lambda st: st.name, SearchType))))
+        return search_type
+
+    @pydantic.validator('criteria')
+    def criteria_must_match_format_for_type(cls, criteria, values):  # pylint:disable=no-self-argument
+        if 'type' not in values:
+            return criteria
+
+        if values['type'] == SearchType.REGISTRATION_NUMBER.value:
```
jguertin, line 33: Note: I limited the effort on this to Base Registration search

bryan-gilbert (approved):

ppr-api/src/endpoints/search.py
```diff
@@ -43,3 +43,12 @@ def read_search(search_id: int, search_repository: repository.search_repository.
     if search_db_row is None:
         raise fastapi.HTTPException(status_code=404, detail="Search record not found")
     return search_db_row
+
+
+@router.post("/searches", response_model=schemas.search.Search, response_model_by_alias=False)
```
WalterMoar: I wish we could lint on quotes - we're going single, yes?

ppr-api/tests/unit/schemas/test_search_schema.py
```diff
@@ -0,0 +1,61 @@
...
+import schemas.search
+
+
+def test_validate_criteria_no_value_with_regnum():
+    try:
+        schemas.search.SearchBase(criteria={}, type=schemas.search.SearchType.REGISTRATION_NUMBER.name)
+    except ValueError:
+        pass
+    else:
+        pytest.fail("A validation error was expected since there was no value field in the criteria")
```
WalterMoar: Quotes again.

ppr-api/src/schemas/search.py
```diff
@@ -1,14 +1,47 @@
...
+    REGISTRATION_NUMBER = 'REGISTRATION_NUMBER'
+    SERIAL_NUMBER = 'SERIAL_NUMBER'
+
+
+class SearchBase(pydantic.BaseModel):
     type: str
     criteria: dict
 
+    @pydantic.validator('type')
+    def type_must_match_search_type(cls, search_type):  # pylint:disable=no-self-argument
```
WalterMoar, line 21: Will the @classmethod annotation prevent the pylint problem?

ppr-api/src/endpoints/search.py
jguertin: Yes :(.  I checked into that yesterday, there's a couple of open source tools: `pylint-quotes` and `flake8-quotes`.  Not sure how comfortable I am bringing them in though.  I'm getting better...  nice catch.

WalterMoar (changes requested):
Some double quotes snuck in.

ppr-api/src/schemas/search.py
jguertin: @validator should be treated as a class method, so adding that decorator would be an equivalent code smell, I think.  I'll add an issue to look into this for tech debt, it that works.

See https://prompt-build--pydantic-docs.netlify.com/#validators for more info on validators

jguertin: Created issue #288

WalterMoar (approved):

gh2os (approved):

Issue #234, opened 2020-01-09
Auth - use in PPR UI


bryan-gilbert:
After we stand up the new dev end point 
- [] update the configuration files used by the auth methods
- [] setup to redirect to auth system on all pages except for the Home page. (This Home page is a throw away pending solution for how users access PPR)
- [] pass the JWT to the API

Expect the above can be done in less than a day.

bryan-gilbert:
PR #234 resolves

PR #282, opened 2020-01-21
If user is not authenticated Issue #234

bryan-gilbert:
and visits one of the PPR pages that requires authentication then redirect to the auth pages.
No return URL is part of the auth system
Fix unit tests now that auth checking is in place; remove console out

### DIFF
--- a/ppr-ui/public/config/configuration.json
+++ b/ppr-ui/public/config/configuration.json
@@ -1,7 +1,7 @@
 {
   "API_URL": "http://localhost:8080/",
   "AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/",
-  "AUTH_URL": "http://localhost:8000/auth/",
+  "AUTH_URL": "https://dev.bcregistry.ca/cooperatives/auth",
   "LAUNCH_DARKLY_CLIENT_KEY": "Client side ID from https://app.launchdarkly.com/settings/projects",
   "PAY_API_URL": "https://pay-api-test.pathfinder.gov.bc.ca/api/v1/",
   "SENTRY_DSN": "",
--- a/ppr-ui/src/router/router.ts
+++ b/ppr-ui/src/router/router.ts
@@ -19,13 +19,12 @@ const router = new VueRouter({
   }
 })
 
-const ENABLED = false
+const ENABLED = true
 
 // if there is no saved Keycloak token, redirect to Auth URL
 router.afterEach((to): void => {
   if (ENABLED) {
     try {
-      console.debug('Router afterEach', to.matched)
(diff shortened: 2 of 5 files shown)


ppr-ui/src/router/router.ts
```diff
@@ -19,13 +19,12 @@ const router = new VueRouter({
   }
 })
 
-const ENABLED = false
+const ENABLED = true
```
WalterMoar: If this is always going to be enabled, can the logic be removed?

ppr-ui/src/views/Home.vue
```diff
@@ -24,6 +24,13 @@
...
                 <h2>Sample home page content</h2>
               </header>
+              <div>Copy needed here to welcome the user to the PPR application.</div>
+            </section>
+            <section v-if="userIsAuthed">
+              <header>
+                <h2>Title needed to welcome authorized user</h2>
+              </header>
+              <div>Copy needed here to welcome the authorized user to the PPR application.</div>
             </section>
```
WalterMoar, line 34: Won't all our users be authenticated?

bryan-gilbert: After consulting with Liv the answer is we will likely, yet to be finalized, have content that is reachable by unauthenticated users

ppr-ui/src/router/router.ts
bryan-gilbert: thanks ... removed.

ppr-ui/public/config/configuration.json
```diff
@@ -1,7 +1,7 @@
 {
   "API_URL": "http://localhost:8080/",
   "AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/",
-  "AUTH_URL": "http://localhost:8000/auth/",
+  "AUTH_URL": "https://dev.bcregistry.ca/cooperatives/auth",
```
jguertin, line 4: Any thoughts on how we can make auth work on our localhosts?

jguertin (approved):

ppr-ui/public/config/configuration.json
WalterMoar: Pasting in a valid JWT was the method used by the other teams. Sub-optimal, for sure.

ppr-ui/tests/unit/SearchSerial.spec.ts
```diff
@@ -82,6 +82,7 @@ describe('SearchSerial.vue', (): void => {
       }
     })
     vm = wrapper.vm
+    sessionStorage.setItem('KEYCLOAK_TOKEN','foo')
```
WalterMoar, line 85: Will get the linting files in today, so that these don't get in.

WalterMoar (approved):

PR #284, opened 2020-01-21
Add an endpoint for search results that returns dummy data

jguertin:
Create the basic structure for `GET /searches/{id}/results` and add a reachable endpoint.  For the moment it contains dummy data, but will have it query the db in future commits

### DIFF
--- a/ppr-api/src/endpoints/search.py
+++ b/ppr-api/src/endpoints/search.py
@@ -1,12 +1,15 @@
 """ Define the endpoints for searching. """
 
+import datetime
 import json
+import typing
 
 import fastapi
 import requests
 from starlette import responses, status
 
 import config
+import schemas.financing_statement
 import schemas.search
 import repository.search_repository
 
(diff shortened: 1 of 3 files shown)


ppr-api/src/endpoints/search.py
```diff
@@ -43,3 +46,24 @@ def read_search(search_id: int, search_repository: repository.search_repository.
     if search_db_row is None:
         raise fastapi.HTTPException(status_code=404, detail="Search record not found")
     return search_db_row
+
+
+@router.get("/searches/{search_id}/results", response_model=typing.List[schemas.search.SearchResult],
```
WalterMoar: Double quotes.

WalterMoar (changes requested):
Just some double quotes, otherwise looks good.

ppr-api/src/endpoints/search.py
bryan-gilbert: Does the lint checker catch this?

WalterMoar: No, but Justin identified some packages that might work. We'll have the same problem in the front end, with the added bonus that in a SFC we'll need doubles in the template and css, and singles in the script 😮

bryan-gilbert (approved):

ppr-api/src/endpoints/search.py
jguertin: Corrected with the rebase

WalterMoar (approved):
Woohoo.

PR #285, opened 2020-01-21
Add a migration to generate the search_result table

jguertin:
Created a migration to generate a search_results table.  This is still lacking a foreign key to the registrations themselves, which has its definition in progress.

### DIFF
--- a/ppr-api/migrations/versions/1461b03c20de_create_search_result_table.py
+++ b/ppr-api/migrations/versions/1461b03c20de_create_search_result_table.py
@@ -1,0 +1,31 @@
+"""Create search_result table
+
+Revision ID: 1461b03c20de
+Revises: c17a9c638a89
+Create Date: 2020-01-16 16:39:55.238345
+
+"""
+from alembic import op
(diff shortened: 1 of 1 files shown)


WalterMoar (approved):

bryan-gilbert (approved):
Is there a table per search type? Should the table name reflect the type?

At this time the registration number is 7 chars. Is this a future requirement or a left over from the legacy system that is no longer important.  In other words, when (not if) will this field need to go to 8 chars? or optionally be a guid?

gh2os (approved):

jguertin:
> Is there a table per search type? Should the table name reflect the type?
> 
> At this time the registration number is 7 chars. Is this a future requirement or a left over from the legacy system that is no longer important. In other words, when (not if) will this field need to go to 8 chars? or optionally be a guid?

No, there is not a table per search type.

Regarding the character width, for today the restriction is on 7 characters.  I'm certain that will change, but we're not clear which direction it will go.  So leaving it as 7 characters for now, and we can alter in the future when that's clearer.

PR #286, opened 2020-01-21
Added SearchInput component with props and events

WalterMoar:
Added a generic search input component that uses properties and events. It can be used for any of our single-field searches. There are some TODO items with associated issues, and an upcoming BUG too, but I wanted to get this merged in.

Note that there are a lot of formatting changes, but we'll get that under control soon I hope.

### DIFF
--- a/ppr-ui/src/search/search-serial.ts
+++ b/ppr-ui/src/search/search-serial.ts
@@ -1,11 +1,12 @@
-import {inject, provide, reactive} from "@vue/composition-api"
-import axiosAuth from "@/utils/axios-auth"
+import { inject, provide, reactive } from '@vue/composition-api'
+import axiosAuth from '@/utils/axios-auth'
 
 // export text constants for use in tests
 export const SS_TEXT = {
   errorMsg: (text): string => `Search serial number error - ${text}`,
   describeValidSerial: 'Serial number can have up to 25 letters or digits',
   defineValid: 'Serial number must be up to 25 letters or digits',
+  label: 'Serial number',
   required: 'Require serial number',
   tooLong: 'Serial number can only have 25 characters'
 }
@@ -22,7 +23,6 @@ interface VFunction {
 }
 
 export class SearcherSerial {
-
   private static _instance: SearcherSerial
 
   public static Instance(baseUrl): SearcherSerial {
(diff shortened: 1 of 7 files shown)


ppr-ui/tests/unit/search/SearchInput.spec.ts
```diff
@@ -0,0 +1,43 @@
+import Vue from 'vue'
+import VueCompositionApi from '@vue/composition-api'
+// import { ref } from '@vue/composition-api'
```
bryan-gilbert, line 3: Do we need this comment

ppr-ui/tests/unit/SearchSerial.spec.ts
```diff
@@ -20,9 +23,8 @@ const featureFlags = FeatureFlags.Instance
 const loadIndicator = LI.Instance
 const search = SI.Instance('http://sample.base.url/')
 
-Vue.use(Vuetify)
-Vue.use(VueCompositionApi)
-
+/* TODO: failing due to setup, and not sure what it is testing. Fix later (https://github.com/bcgov/ppr/issues/283)
+import SearchPage from '@/views/SearchPage.vue'
```
bryan-gilbert, line 27: You can also do a describe.skip instead of using a comment block.

ppr-ui/src/search/search-serial.ts
```diff
@@ -1,11 +1,12 @@
-import {inject, provide, reactive} from "@vue/composition-api"
-import axiosAuth from "@/utils/axios-auth"
+import { inject, provide, reactive } from '@vue/composition-api'
```
bryan-gilbert, line 1: This commit has a lot of whitespace linting changes.  Why are these appearing? Have the linting rules been tightened up?  (OK with me)  Asking if the Github lint checking action will catch future changes like this.

ppr-ui/tests/unit/SearchSerial.spec.ts
WalterMoar: Ooh, awesome - thanks! Will use it next time.

ppr-ui/src/search/search-serial.ts
WalterMoar: It's the linter in VS Code. I'll commit the files to the project so that we're hopefully all on the same page. Will try to get it done today.

ppr-ui/tests/unit/search/SearchInput.spec.ts
WalterMoar: It's for the commented out tests - had I used describe.skip it would have been a lot cleaner.

bryan-gilbert (approved):

ppr-ui/src/search/search-serial.ts
jguertin: Our lint settings run during tests must not be configured to the same level of spacing as the formatter in VS Code.  Otherwise this would have caused check errors.

jguertin (approved):

PR #343, opened 2020-01-23
To be reactive the property needs to declared and initialized

bryan-gilbert:


WalterMoar (changes requested):
Please remove the two console.log statements.

WalterMoar:
Can you connect this to the issue?

ppr-ui/src/search/search-serial.ts
```diff
@@ -39,7 +39,10 @@ export class SearcherSerial {
 
   private readonly _serialNumberRules: VFunction[] // ((v: string) => (string | boolean))[]
 
-  public constructor(baseUrl) {
+  public constructor(baseUrl: string) {
+    this._errorMessage = ''
+    this._results = []
+    this._term = ''
```
WalterMoar, line 45: The label should be reactive as well.

bryan-gilbert: the label contains a static value and does not exist as a private property.

WalterMoar: Oh yeah - good point.

WalterMoar (approved):

PR #344, opened 2020-01-23
Auth does not work on local dev

bryan-gilbert:
As a developer it is essential to run the app locally. Yet auth is, by design, tied to the bcreg domain.  

Developers need to disable auth on their local machine.

This workaround introduces risk that someone may check in code with this flag set to true.

Reviewers .. please discuss options.

### DIFF
--- a/ppr-ui/src/router/router.ts
+++ b/ppr-ui/src/router/router.ts
@@ -19,14 +19,20 @@ const router = new VueRouter({
   }
 })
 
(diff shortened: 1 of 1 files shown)


WalterMoar (changes requested):
Let's discuss, I don't think we want to do it this way.

bryan-gilbert:
Will find another way

PR #367, opened 2020-01-27
Feature flags for document registration number.

WalterMoar:
Re-purposed one of the PoC feature flags for the Document Registration Number search. Removed broken tests until we determine what we want to unit test for the flags.


ppr-ui/src/flags/ld-client.ts
```diff
@@ -27,10 +27,10 @@ class VueLdClient {
     }
   }
 
-  private updateFlags(allFlags): void {
+  private updateFlags(allFlags: LDFlagSet): void {
     const flags = FeatureFlags.Instance
     console.debug('LDFlags on change flags', allFlags)
-    flags.feature1 = allFlags['poc-feature-1']
+    flags.feature1 = allFlags['search-document-registration-number']
```
jguertin, line 33: Will you also rename `feature1` to reflect to the intent?

jguertin (approved):
looks good, 1 question, non-blocking

ppr-ui/src/flags/ld-client.ts
WalterMoar: Yes indeed! We also need to something more scalable for the flags, but might not get to that for this sprint.

ppr-ui/src/views/Home.vue
```diff
@@ -82,25 +70,28 @@
...
+
+    const userIsAuthed = computed((): boolean => !!sessionStorage.getItem('KEYCLOAK_TOKEN'))
+    const userCanSearch = computed((): boolean => flags.feature1 && userIsAuthed.value)
+
     function goSearch(): void {
       router.push('search')
     }
 
-    return { flags, feature2, featureOneLabel, featureTwoLabel, goSearch, userIsAuthed }
+    return { flags, featureOneLabel, featureTwoLabel, goSearch, userCanSearch, userIsAuthed }
```
bryan-gilbert: Can also remove the two featureXXXLabel computes.

bryan-gilbert (approved):
one non-blocking change suggested

ppr-ui/src/views/Home.vue
WalterMoar: Yes! I completely missed that.

ppr-ui/src/flags/ld-client.ts
WalterMoar: https://github.com/bcgov/ppr/issues/369

PR #370, opened 2020-01-27
Renamed configuration.json file

WalterMoar:
Renamed this file so I'm not constantly editing and reverting it. It gets overwritten by a configmap when deployed to the cluster.

Also removed the other sample file.


bryan-gilbert (approved):
Good to have this file.  
Let's update the .gitignore file too. Change .configuration.json
 to just configuration.json
to prevent checking in the file we use on our dev boxwes.

WalterMoar:
> to prevent checking in the file we use on our dev boxwes.

Oh right!

jguertin (approved):

PR #372, opened 2020-01-27
Update copyright year

WalterMoar:
Reformatted to 120 columns, this was created when we were doing 80.


WalterMoar:
Nevermind, I have been bamboozled. We should not update copyright dates, the original date is what matters.

PR #375, opened 2020-01-28
Api search by registration number

jguertin:
Update registration number searches to run against the registrations.  This includes changes to:
- Read data from the tables in the database
- A migration script that creates a foreign key relationship between search_result and registration
- Updates to previous migrations to provide some more leniency and clean up column types
- Updates to existing tests to make them passing

### DIFF
--- a/ppr-api/migrations/versions/947ffeeeda60_create_foreign_key_relationship_between_.py
+++ b/ppr-api/migrations/versions/947ffeeeda60_create_foreign_key_relationship_between_.py
@@ -1,0 +1,28 @@
+"""Create foreign key relationship between search_result and registration
+
+Revision ID: 947ffeeeda60
+Revises: 1461b03c20de
+Create Date: 2020-01-24 16:07:20.288676
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '947ffeeeda60'
+down_revision = 'fc3c53132a9b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Remove orphaned search_result rows (don't not have a valid registration_number)
+    op.execute('DELETE FROM search_result sr WHERE NOT EXISTS '
+               '(SELECT 1 FROM registration r WHERE r.reg_number = sr.registration_number)')
+
+    op.create_foreign_key('search_result_registration_fk', 'search_result', 'registration', ['registration_number'],
+                          ['reg_number'])
+
+
+def downgrade():
+    op.drop_constraint('search_result_registration_fk', 'search_result')
--- a/ppr-api/src/models/financing_statement.py
+++ b/ppr-api/src/models/financing_statement.py
@@ -14,7 +14,7 @@ class FinancingStatement(BaseORM):
     life_in_years = sqlalchemy.Column('life', sqlalchemy.Integer)
     expiry_date = sqlalchemy.Column(sqlalchemy.Date)
     discharged = sqlalchemy.Column(sqlalchemy.BOOLEAN)
-    last_updated = sqlalchemy.Column('last_update_timestamp', sqlalchemy.DateTime)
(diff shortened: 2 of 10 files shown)


WalterMoar (approved):

gh2os (approved):

ppr-api/migrations/versions/947ffeeeda60_create_foreign_key_relationship_between_.py
```diff
@@ -0,0 +1,28 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '947ffeeeda60'
+down_revision = 'fc3c53132a9b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Remove orphaned search_result rows (don't not have a valid registration_number)
```
bryan-gilbert, line 19: double negative intended?  "don't not" ==> "does"?

ppr-api/src/models/financing_statement.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class FinancingStatement(BaseORM):
+    __tablename__ = 'financing_statement'
+
+    registration_number = sqlalchemy.Column('reg_number', sqlalchemy.String(length=10), primary_key=True)
+    registration_type_code = sqlalchemy.Column('reg_type_cd', sqlalchemy.CHAR(length=2))
+    status = sqlalchemy.Column(sqlalchemy.CHAR(length=1))
+    life_in_years = sqlalchemy.Column('life', sqlalchemy.Integer)
+    expiry_date = sqlalchemy.Column(sqlalchemy.Date)
```
bryan-gilbert, line 15: Are these two values consistent with each other? Just curious.  For example, the FS will have a creation date.  It'll have an expiry date based on what the user pays for.  Will "life in years" be the calculated number of years between these two dates or can this life value have a number that is set some other way?

ppr-api/src/models/financing_statement.py
```diff
@@ -0,0 +1,34 @@
...
+    discharged = sqlalchemy.Column(sqlalchemy.BOOLEAN)
```
bryan-gilbert, line 16: If discharged is true will life in years be zeroed ... will expiry date be non-empty?  maybe there is no answer just yet.

ppr-api/src/schemas/search.py
```diff
@@ -42,8 +42,6 @@ def criteria_must_match_format_for_type(cls, criteria, values):  # pylint:disabl
                 raise ValueError('"value" is required in criteria')
             if not criteria['value'].isalnum():
                 raise ValueError('"value" must be alphanumeric')
-            if len(criteria['value']) != 7:
-                raise ValueError('"value" must be 7 characters')
```
bryan-gilbert, line 46: so the api no longer checks for 7 characters?

bryan-gilbert (approved):

ppr-api/src/models/financing_statement.py
jguertin: I think that is the intent, `expiryDate` will need to be calculated base on life.  I don't love it, but it's what it is.

ppr-api/src/models/financing_statement.py
jguertin: no, they are not mutually exclusive

ppr-api/src/schemas/search.py
jguertin: Correct, for the time being.  numbers created in the new system will be 8 digits.  I could've enforced 7 or 8, but figured it may not be needed.  We may need to actually remove the `.isalnum()` check, as well.

ppr-api/migrations/versions/947ffeeeda60_create_foreign_key_relationship_between_.py
jguertin: Nice catch.  I'll fix it in another commit.

Issue #379, opened 2020-01-28
 ppr-ui: create search reg num module


bryan-gilbert:
Create a Vue composition api module to support the registration number query, validation, and fetch.  This module supports input validation, error message generation, api post and get calls and provides the search results.  This module is injected into the container that has the reg num search input and the container that will have the reg num search results.

Issue #374, opened 2020-01-28
ppr-ui: use less strict typing in SearchInput unit tests


WalterMoar:
For some reason the SearchInput unit tests are failing due to type errors. It works fine when doing `npm run test:unit` or in the GitHub Actions, but `npm run serve` for some reason runs the tests and produces the errors.

Since we do no require strict types with the Composition API framework testing, we can switch all instances of `Wrapper<SearchInput>` to `Wrapper<Vue>`.

WalterMoar:
Closing this, the work was done as part of another ticket.

PR #380, opened 2020-01-29
#379 Reg number search module

bryan-gilbert:


bryan-gilbert:
the code smell is a good catch can be fixed in the next commit which is related to this one

ppr-ui/src/App.vue
```diff
@@ -61,6 +62,7 @@ export default createComponent({
     provideFeatureFlags()
     provideLoadIndicator()
     provideSearcherSerial(AppData.config.pprApiUrl)
+    provideSearcherRegNum(AppData.config.pprApiUrl)
```
WalterMoar, line 65: I think we need to re-evaluate this provide/inject pattern before next sprint. The Vue documentation says "_provide and inject are primarily provided for advanced plugin / component library use cases. It is NOT recommended to use them in generic application code._" Perhaps this changes with the Composition API but it looks overly complicated.

ppr-ui/src/search/search-regnum.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  /*
+  Private members for holding the search term, results and/or error
+   */
+  private _errorMessage: string
+  private _results: RegNumResult[]
+  private _term: string
+  private _searchId: string
+  private readonly _baseUrl: string
+  private readonly _serialNumberRules: VFunction[]
```
jguertin: `_serialNumberRules` feels like a bit of a misnomer.

ppr-ui/src/App.vue
bryan-gilbert: The provide and inject they are referring to is the Vue 2.0 provide and inject and the comment is absolutely true.  That provide and inject mechanism was non-reactive and its use could easily lead to errors as developers may have expected it to be reactive. Hence the advice to use it for plugin injection.   
The Composition Api depends on the new "reactive" provides and injects.  I wonder if the Vue community is forming helpful ways of calling these two approaches by different names.
Nonetheless we can certainly review if we wish to move to Vue 3.0 approaches.

ppr-ui/src/search/search-regnum.ts
```diff
@@ -0,0 +1,190 @@
...
+import axiosAuth from '@/utils/axios-auth'
+import {AxiosResponse} from 'axios'
+
+const MAX_CHAR = 8
+const REG_EXP = /^[0-9a-zA-Z]{1,8}$/
+
+// export text constants for use in tests
+export const SS_TEXT = {
+  errorMsg: (text: string): string => `Search document registration number error - ${text}`,
+  describeValid: `Document registration number has any combination of letters, characters, and numbers (up to ${MAX_CHAR} characters)`,
```
WalterMoar: "characters"? Isn't it just letters and numbers?

WalterMoar: And are we using "**Document** registration number"?

ppr-ui/src/search/search-regnum.ts
```diff
@@ -0,0 +1,190 @@
...
+      },
+      (value): (boolean | string) => {
+        return REG_EXP.test(value) || SS_TEXT.describeValid
+      }
+    ]
+  }
+
+  public get errorMessage(): string {
+    return this._errorMessage
+  }
```
WalterMoar, line 75: I do find this much easier to read than the single-line equivalent 👍

ppr-ui/src/App.vue
jguertin, line 65: I'm not clear on the reason we're providing the api url as a "SearcherRegNum".

ppr-ui/src/search/search-regnum.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  // This is a public method is for testing the validation rules.
+  public validate(input: string):
+  string | undefined {
+    // iterate over the rules ... if any rule results in a string error message ...
+    // ... then return the result of running that rule.
+    let rule = this._serialNumberRules.find((rule: VFunction): boolean => typeof rule(input) === 'string')
+    return rule ? rule(input) as string : undefined
+  }
+
```
WalterMoar: Is this only used for testing? Can it be moved to the tests?

ppr-ui/src/search/search-regnum.ts
bryan-gilbert: On the glass, I really don't know the answer.  This maybe something our UX team can look at later?

ppr-ui/src/search/search-regnum.ts
```diff
@@ -0,0 +1,190 @@
...
+        "debtors": [],
+        "vehicleCollateral": [],
+        "generalCollateral": [],
+        "expiryDate": null,
+        "baseRegistrationNumber": "1234567",
+        "documentId": "B9876543",
+        "registrationDateTime": "2020-01-27T10:55:09.530526"
+      }
+    }
+*/
```
WalterMoar: Can this be deleted? We already have a lot of this to clean up.

ppr-ui/src/views/SearchPage.vue
```diff
@@ -46,8 +54,8 @@ export default createComponent({
   setup(): Data {
     const loadIndicator = useLoadIndicator()
     const { router } = useRouter()
-
     const searcherSerial = useSearcherSerial()
+    const searchRegNum = useSearcherRegNum()
```
WalterMoar: We should be consistent in naming, if possible.

ppr-ui/src/search/search-regnum.ts
bryan-gilbert: 50/50 for me.  This method also means this module can be used with a non Vuetify library.  The Vuetify input takes a set of validation rules and, in effect, runs this method.  The code should exist somewhere.  If we leave it here then it's ready as useful module API.  
Where do you wish it?

ppr-ui/tests/unit/search/SearchInput.spec.ts
```diff
@@ -12,7 +12,7 @@ const vuetify = new Vuetify()
 
 describe('SearchInput.vue', (): void => {
   it('handles no props', (): void => {
-    const wrapper: Wrapper<SearchInput> = mount(SearchInput, { vuetify })
+    const wrapper: Wrapper<Vue> = mount(SearchInput, { vuetify })
```
WalterMoar, line 15: This is a different Issue (https://github.com/bcgov/ppr/issues/374) which isn't part of this sprint and isn't anything to do with this Issue.

ppr-ui/tests/unit/search/search-regnum.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+import axios from '@/utils/axios-auth'
+
+jest.mock('axios')
+const mockedAxios = axios as jest.Mocked<typeof axios>
+
+Vue.use(VueCompositionApi)
+
+// TODO find a way to test the provides and injects see https://github.com/bcgov/ppr/issues/377
+describe.skip('provide inject', (): void => {
+  it('Doc reg search module provides and injects SearcherDocReg', (): void => {
```
WalterMoar: If we're going to drop the "Doc" we need to be consistent.

ppr-ui/src/search/search-regnum.ts
bryan-gilbert: I need this until I build the results component.

ppr-ui/src/views/SearchPage.vue
bryan-gilbert: ok ... will submit an update in a sec

ppr-ui/src/search/search-regnum.ts
jguertin: `Document` is confusing, would love to remove it from our vocabulary until told otherwise.  ;)

ppr-ui/src/search/search-regnum.ts
```diff
@@ -0,0 +1,190 @@
+import {inject, provide, reactive} from '@vue/composition-api'
```
jguertin, line 1: I'm concerned that there's a few mixed purposes in here, and this file is trying to do more than it should.  @WalterMoar @bryan-gilbert Can we discuss this in the morning?

jguertin (changes requested):

jguertin (approved):
Approved as discussed, to move forward with sprint goals

WalterMoar (approved):

Issue #355, opened 2020-01-24
ppr api and db: Write alembic scripts to build v1 financing data model


jguertin:
As an output of entity#2263, we'll receive our first crack at the target data model for the new PPR system.  Once this is ready, write alembic migration scripts to build the tables in the postgres database.

jguertin:
Created financing statement document and code tables: #366

gh2os:
Moving back to in progress to create FS Party tables

PR #381, opened 2020-01-29
Alembic Scripts for Address Migration #355

gh2os:
Alembic script that creates the address, provincial state type, and country type tables.


ppr-api/migrations/versions/6411b673d98b_create_address_tables.py
```diff
@@ -0,0 +1,389 @@
...
+        {'country_type_cd': 'YU', 'description': 'YUGOSLAVIA'},
+        {'country_type_cd': 'ZA', 'description': 'SOUTH AFRICA'},
+        {'country_type_cd': 'ZM', 'description': 'ZAMBIA'},
+        {'country_type_cd': 'ZW', 'description': 'ZIMBABWE'}
+    ])
+
+    prov_state_type_table = op.create_table(
+        'prov_state_type',
+        sa.Column('province', sa.CHAR(length=2), primary_key=True),
+        sa.Column('country_type_cd', sa.CHAR(length=2), sa.ForeignKey('country_type.country_type_cd')),
```
jguertin, line 287: province code is unique within CA and US, just wondering if country_type_cd should be included as a composite PK?  Theoretically yes, but maybe cross that bridge when we come to it?

ppr-api/migrations/versions/6411b673d98b_create_address_tables.py
```diff
@@ -0,0 +1,389 @@
...
+def upgrade():
+
+    country_type_table = op.create_table(
+        'country_type',
+        sa.Column('country_type_cd', sa.CHAR(length=2), primary_key=True),
+        sa.Column('description', sa.String(length=75))
+    )
+
+    op.bulk_insert(country_type_table, [
+        {'country_type_cd': 'AD', 'description': 'ANDORRA'},
```
jguertin: Just wondering about case for description.  All upper makes sense in the IMS system, wondering if it's appropriate going forward? It's a lot of effort to change, so maybe leave as is for now?

jguertin (approved):
Looks good, a couple of questions for consideration, but not blocking.

ppr-api/migrations/versions/6411b673d98b_create_address_tables.py
gh2os: Ah, I see what you mean. Happy to cross the bridge when we get to it, or we can look into addressing that now. I am indifferent.

WalterMoar (approved):

jguertin (approved):

PR #382, opened 2020-01-29
346: Create Test Strategy

rstens:
The first submission of the PPR Test Strategy


jguertin (approved):
As far as location for the PPR test strategy goes, it seems like as good a place as any to put it.  As for the content, I find the file to be quite long and quite broad.  Who's the target audience for this document?  Is there a summary of the proposed approach which speaks to what you suggest our team should be doing?

Approving for location, but I think it would be good to develop consensus amongst team members on actual approach.

rstens:
Yes, consensus is good, I'll prepare a condensed overview for discussion.

WalterMoar:
Seems like a good location for it.

PR #383, opened 2020-01-29
RegNum search results

bryan-gilbert:
Provide a component to show the financing statements that are retrieved via a registration number search.  This is, purposely, meant to be a code like display as we are developing while UX research is underway.


ppr-ui/src/search/SearchResultsRegistrationNumber.vue
```diff
@@ -0,0 +1,38 @@
...
+    </p>
+    <div>
+      <pre>{{ results }}</pre>
+    </div>
+  </section>
+</template>
+
+<script lang="ts">
+import {computed, createComponent} from "@vue/composition-api"
+import {Data} from "@vue/composition-api/dist/ts-api/component"
```
WalterMoar: Your VS Code still isn't formatting code. Also, we're using single quotes in TypeScript.

ppr-ui/src/search/SearchResultsRegistrationNumber.vue
```diff
@@ -0,0 +1,38 @@
+<template>
```
WalterMoar: I don't think we want a component for the results from every type of search. I know this is temporary, but it should at least be leaning in the direction of being a reusable component.

ppr-ui/src/search/SearchResultsRegistrationNumber.vue
bryan-gilbert: Help.  I used VS code this time.

ppr-ui/src/search/SearchResultsRegistrationNumber.vue
WalterMoar: ^^^ Worst, vague, rushed, comment ever.

bryan-gilbert: but thankfully we can chat ... hope the new commit helps!

WalterMoar (approved):

PR #395, opened 2020-02-01
244-Sentry in PPR UI

bryan-gilbert:
Update how we setup and use Sentry in the UI.
Note that all console.error messages as well as all thrown errors will go to Sentry


ppr-ui/src/views/SearchPage.vue
```diff
@@ -64,8 +63,9 @@ export default createComponent({
         .then((): void => {
           router.push('results')
         })
-        .catch((/* errorMessage: string */): void => {
-          // TODO: log this somewhere, probably Sentry. https://github.com/bcgov/ppr/issues/244
+        .catch((errorMessage: string): void => {
```
jguertin, line 64: Doesn't catch usually receive an object of type `Error`, rather than `string`?

ppr-ui/src/sentry-helper.ts
```diff
@@ -0,0 +1,74 @@
...
+            // Vue, this is optional assuming window.Vue is present
+            logErrors: true, // call original Vue logError as well
+            attachProps: true // send Vue props to sentry
+          }),
+          /*
+          CaptureConsole
+          This integration captures all Console API calls and redirects them to Sentry using 
+          captureMessage call. It then retriggers to preserve default native behaviour.
+          */
+          new Integrations.CaptureConsole({
```
WalterMoar: Nice! 👍

ppr-ui/src/views/SearchPage.vue
WalterMoar: Oh, that's my fault - I think the type is `any`. You can throw a string, which is what is happening here, but yes the type actually should be `any` and not `string`.

WalterMoar (changes requested):
We have the other "helpers" in the utils directory, can sentry-helper.ts be moved there too?

That and the type problem Justin pointed out (my fault!) are all that I think it needs.

ppr-ui/src/views/SearchPage.vue
bryan-gilbert: No this is my fault in the way I set up the api calls.  My initial implementation is too complex and it does resolve and reject with a string.  In this case, the catch is getting a string.  The next time I touch these api calls, and now that we have sentry in place, I'll simplify the calls to a single line. (Follow the pattern used in sbc_auth for api calls).

bryan-gilbert:
> We have the other "helpers" in the utils directory, can sentry-helper.ts be moved there too? 
> That and the type problem Justin pointed out (my fault!) are all that I think it needs.

Good idea to group this with the others.  Thanks

WalterMoar (approved):

PR #396, opened 2020-02-01
Add an authentication module which calls the Auth API with the provided token

jguertin:
This can be leveraged by adding the following parameter to each endpoint that requires auth: `user: auth.authentication.User = fastapi.Depends(auth.authentication.get_current_user)`

I've left this portion out for the moment as it will break integration tests, since the tests do not have a valid token when running, so can't use the actual API.  I'll work on that next week.

### DIFF
--- a/ppr-api/src/config.py
+++ b/ppr-api/src/config.py
@@ -12,6 +12,8 @@ import dotenv
 
 dotenv.load_dotenv()
 
+AUTH_API_URL = os.getenv("PPR_API_AUTH_API_URL")
+
 CORS_ORIGINS = os.getenv('PPR_API_ALLOWED_ORIGINS', 'http://localhost:8080 http://localhost:8081').split()
 
 DB_HOSTNAME = os.getenv('PPR_API_DB_HOSTNAME')
--- a/ppr-api/tests/unit/auth/test_authentication.py
+++ b/ppr-api/tests/unit/auth/test_authentication.py
@@ -1,0 +1,78 @@
+import http
+from unittest.mock import patch
+
+import fastapi.security.http
+import pytest
+import requests
+
+import auth.authentication
+
+
+@patch('requests.get')
+def test_get_user_from_auth_with_successful_auth(mock_get):
+    mock_get.return_value = requests.Response()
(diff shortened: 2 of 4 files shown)


ppr-api/src/config.py
```diff
@@ -12,6 +12,8 @@
 
 dotenv.load_dotenv()
 
+AUTH_API_URL = os.getenv("PPR_API_AUTH_API_URL")
```
jguertin, line 15: @WalterMoar Thoughts on the `PPR_API_` prefix?  I used it to be consistent with other variable, but wondering whether it makes sense in this case.

ppr-api/tests/unit/auth/test_authentication.py
```diff
@@ -0,0 +1,78 @@
...
+import requests
+
+import auth.authentication
+
+
+@patch('requests.get')
+def test_get_user_from_auth_with_successful_auth(mock_get):
+    mock_get.return_value = requests.Response()
+    mock_get.return_value.status_code = http.HTTPStatus.OK
+    mock_get.return_value._content = b'{"test": "json"}'
```
gh2os, line 15: Can you explain to me why this and the other "_contents" (lines 27, 43, and 59) are converted to bytes? Just for my own knowledge

gh2os (approved):

ppr-api/tests/unit/auth/test_authentication.py
jguertin: That is for the internal workings of the `requests.Response` class, which stores the response body internally as bytes.

ppr-api/src/config.py
WalterMoar: I was using it solely because in other projects where we had multiple python APIs, having the .env file higher up the directory tree would cause contention. I'm OK with dropping it, as long as we drop it everywhere.

WalterMoar (approved):

ppr-api/src/config.py
jguertin: I'll leave it for now, then.  Thanks.

PR #402, opened 2020-02-03
Updated import of sentry-helper

WalterMoar:
The previous change for sentry borked the import.


jguertin (approved):

bryan-gilbert (approved):
Thanks!  Wonder why the build action didn't catch this.  Wonder why the dev (ahem) didn't catch this too!

PR #405, opened 2020-02-03
Convert the API code base to use single quotes consistently.

jguertin:
Includes the addition of `flake8-quotes` extension for flake8.

### DIFF
--- a/.github/workflows/compile-and-test-stage.yml
+++ b/.github/workflows/compile-and-test-stage.yml
@@ -31,7 +31,7 @@ jobs:
         run: |
           pylint -E src
           pylint -E --disable=E1101 migrations
-          flake8 src migrations tests
+          flake8 src migrations tests *.py
       - name: Execute Unit Tests with PyTest
         working-directory: ppr-api
         run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/compile-and-test-stage.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: |
           pylint -E src
           pylint -E --disable=E1101 migrations
-          flake8 src migrations tests
+          flake8 src migrations tests *.py
```
jguertin, line 34: Just adding python files in the PPR directory to this linting

ppr-api/requirements/dev.txt
```diff
@@ -4,6 +4,7 @@
 autopep8
 fastapi==0.42.0 # version tied to https://github.com/tiangolo/uvicorn-gunicorn-fastapi-docker
 flake8==3.7.9
+flake8-quotes==2.1.1
```
jguertin, line 7: What do you think, should we keep this? https://pypi.org/project/flake8-quotes/2.1.1/ https://github.com/zheller/flake8-quotes

By default it enforces single-quotes, and as it extends flake8, no additional configuration is needed unless we move away from the defaults.  That said, I'm not clear how safe the source is, nor how widely used this is.

It does make enforcement easy.

@WalterMoar @gh20s

WalterMoar: I like easy enforcement, and I'm jealous :)

WalterMoar (approved):

gh2os (approved):

ppr-api/requirements/dev.txt
gh2os: I like it as well :)

PR #419, opened 2020-02-05
Formatting of Vue and Typescript files

WalterMoar:
Formatted the Vue files so that we should be free of this in the future. Did some typescript files too. Fixed one typescript strict=true error.


bryan-gilbert (approved):
with appreciation

jguertin (approved):
👍

PR #420, opened 2020-02-05
Fix API Coverage reports

jguertin:
Fix up the PPR API test coverage so it reports on src rather than tests.  Also includes changes so content for both unit and integration tests is sent to codecov.


rstens (approved):
Looks good. This was indeed still an issue.

bryan-gilbert (approved):

WalterMoar (approved):

PR #440, opened 2020-02-06
Test infrastructure

rstens:
The initial submission of the test framework.
It contains the structure for the directories, code, build configs for images and pipelines.


testing/functional/.gitignore
```diff
@@ -0,0 +1,11 @@
+build
```
bryan-gilbert, line 1: Just for information, we've tried to place all ignores in the root gitignore.

bryan-gilbert (approved):
look forward to learning about the details.

PR #444, opened 2020-02-07
Corrected a mispelling of Aircraft in the vehicle type descriptions

jguertin:
Aircraft is 1 word, not 2  🤷‍♂


bryan-gilbert (approved):
such a drag that air craft appears twice .... so close to that single char pr

PR #445, opened 2020-02-10
Remove the POC serial search from the web-ui

jguertin:


ppr-ui/tests/unit/ResultsPage.spec.ts
```diff
@@ -0,0 +1,62 @@
+import Vue from 'vue'
```
jguertin, line 1: This file was actually extracted from `SearchSerial.spec.ts`

WalterMoar (approved):

PR #450, opened 2020-02-10
Set the feature flags user from the JWT

WalterMoar:
Using the unique but rather ugly `username` key in the JWT to use for the key in launch darkly.


ppr-ui/src/main.ts
```diff
@@ -22,17 +23,15 @@ Vue.config.productionTip = false
...
 
 configHelper.fetchConfig()
   .then((cfg: Config): void => {
     appConfig = cfg
     return SentryHelper.setup(appConfig)
-  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
   }).then(() => {
-    return initializeVueLdClient(appConfig.launchDarklyClientKey, userKey)
+    return initializeVueLdClient(appConfig.launchDarklyClientKey, getJwtValue('username'))
```
jguertin, line 34: username works as it will be unique for the user, but I'm not sure if it's a useful value atm.  `sub` is the "Keycloak User ID" which is the current key into the SBC Auth system.  You don't need to change this, it's just FYI in case it informs you.

jguertin (approved):

ppr-ui/src/main.ts
WalterMoar: It's far from useful! I doubt we'll be enabling flags based on BCSC users. However once we get IDIR users in there we can decide on what we want more permanently - I'm guessing flags for staff will be far more common.

bryan-gilbert: does the jwt contain a isStaff indicator?

bryan-gilbert (approved):

ppr-ui/src/main.ts
WalterMoar: The JWT contains roles, and the auth folks would be the ones to set up a staff role when needed.

PR #458, opened 2020-02-11
Add Auth API URL to API configuration

jguertin:
Previous miss on adding configuration

### DIFF
--- a/ppr-api/openshift/ppr-api-dc.yaml
+++ b/ppr-api/openshift/ppr-api-dc.yaml
@@ -41,6 +41,8 @@ objects:
             value: "8080"
           - name: PPR_API_ALLOWED_ORIGINS
             value: ${CORS_ORIGINS}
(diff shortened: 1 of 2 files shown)


bryan-gilbert (approved):

ppr-api/openshift/ppr-api-dc.yaml
```diff
@@ -152,6 +154,11 @@ parameters:
   name: API_NAME
   required: true
   value: ppr-api
+- description: The path to the SBC Auth API.
+  displayName: Auth API URL
+  name: AUTH_API_URL
+  required: truehttps://auth-api-dev.pathfinder.gov.bc.ca/api/v1
+  value:
```
WalterMoar: Looks like the value ended up on the wrong line here.

jguertin: Oh, weird.  Nice catch.

PR #465, opened 2020-02-12
Updated web text for production deployment

WalterMoar:
This isn't final text, but it will do so that we can get deployed to production.

### DIFF
--- a/ppr-ui/public/index.html
+++ b/ppr-ui/public/index.html
@@ -1,24 +1,24 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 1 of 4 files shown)


jguertin (approved):

ppr-ui/public/index.html
```diff
@@ -1,24 +1,24 @@
...
+</head>
+
+<body>
+  <noscript>
+    <strong>Please enable JavaScript to use the Personal Property Registry application.</strong>
+  </noscript>
+  <div id="app"></div>
+</body>
+<script>
+  sessionStorage.setItem('REDIRECTED', 'false')
```
bryan-gilbert: Is this redirect session value still needed?

bryan-gilbert (approved):

ppr-ui/public/index.html
WalterMoar: We don't use it anywhere else. Will remove.

PR #472, opened 2020-02-13
Update to the tests to run against PPR - POC

rstens:
Updating the tests, so they now run against PPR, instead of the cooperatives site.
This is still POC and the tests will change as more functionality is added,


testing/functional/src/test/groovy/specs/StartSpec.groovy
```diff
@@ -0,0 +1,83 @@
...
+            at ResultPage
+            assert resultShown
+        when: "I log out"  
+            to StartPage  
+            logoutuser()
+        then: "I see the cooperatives start page"
+            
+        where: "I use the following data set"
+            Id              || Password ||  GivenName  
+            "BCREG0001"     || "98901"  || "BCREGTEST Dalia"
```
jguertin: Should we be putting these credentials in the source code?

rstens: No, we should not. I'll update.

jguertin (approved):

PR #473, opened 2020-02-13
Create a re-usable deployment script configured for the development environment

jguertin:
For the time being this can be used manually from the command line.  One the image is pushed by GitHub actions after a merge to `master`, you can deploy to dev with:
1. `cd ppr-api/openshift`
2. `./deploy_dev.sh`

I've removed the "configuration trigger" in the DC file, so even though the script gets re-applied, the rollout won't happen until the image is re-tagged.

The next steps are to start automating this process and leverage 1Password.

### DIFF
--- a/ppr-api/openshift/ppr-api-dc.yaml
+++ b/ppr-api/openshift/ppr-api-dc.yaml
@@ -109,7 +109,6 @@ objects:
           name: ${API_NAME}:${IMAGE_TAG}
           namespace: zwmtib-tools
       type: ImageChange
-    - type: ConfigChange
 - apiVersion: v1
   kind: Service
   metadata:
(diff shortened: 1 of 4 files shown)


ppr-api/openshift/ppr-api-dc.yaml
```diff
@@ -169,11 +168,6 @@ parameters:
   name: ENVIRONMENT
   required: true
   value: dev
-- description: The image tag to build from.
-  displayName: Image Tag
-  name: IMAGE_TAG
-  required: true
-  value: dev
```
WalterMoar: This is used on line 109, no?

jguertin: Yes, as discussed, I'll put it back

WalterMoar (approved):

PR #475, opened 2020-02-13
Cleaned up config and added tests

WalterMoar:
Started looking at the test coverage problem and decided to clean up code in addition to writing the tests.


bryan-gilbert (approved):
A good use of lessons learned and a substantial improvement

PR #485, opened 2020-02-14
Reset the application style sheets

bryan-gilbert:
First look at PPR style sheets since taking the first cut from the lear project.
Now have taken the latest from the auth-ui project.
Removes a lot of stuff we don't need.
Resolves the problem of our home page not being responsive.


jguertin (commented):
Nice.

jguertin (approved):

PR #486, opened 2020-02-14
Create payment request on new search in API

jguertin:
When a search is created, send a request to the Payment Service to create a new payment record.  Here's a sample response. Updates to the spec will follow when we're happy with this approach.

This change does not include persisting the payment information, nor returning it in a GET.  The intent for now is to enable the UI to receive a valid payment id so it can handle the workflow correctly.

This is a sample POST response:
```json
{
    "type": "REGISTRATION_NUMBER",
    "criteria": {
        "value": "123456B"
    },
    "id": 35,
    "searchDateTime": "2020-02-14T22:54:14+00:00",
    "payment": {
        "id": 1641,
        "status": "CREATED",
        "method": "CC"
    }
}
```

### DIFF
--- a/ppr-api/src/endpoints/search.py
+++ b/ppr-api/src/endpoints/search.py
@@ -9,6 +9,7 @@ import auth.authentication
 import models.search
 import schemas.financing_statement
 import schemas.search
+import services.payment_service
 import repository.financing_statement_repository
 import repository.search_repository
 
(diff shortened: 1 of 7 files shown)


ppr-api/src/endpoints/search.py
```diff
@@ -48,7 +50,11 @@ def create_search(response: responses.Response, search_input: schemas.search.Sea
 
     search_model = search_repository.create_search(search_input, exact_matches, similar_matches, user)
     response.status_code = status.HTTP_201_CREATED
-    return search_model
+
+    # TODO capture the payment detail in the database for later use
```
bryan-gilbert, line 54: I'm ok with this todo unlinked to any issue because i know we're coming back here really soon

bryan-gilbert (approved):
good for our current needs

PR #529, opened 2020-02-20
Restore sticky footer fix

bryan-gilbert:
After syncing the style sheets with the other teams the footer started to float again.  Placing the fix in the per app styles.scss file rather than one of the other style sheets that may, again, be synced with other teams.
For the original issue see issue 232 - sticky footer #389


WalterMoar:
Let's discuss - I'd like to make sure we're not going to have the same bugs as before.

WalterMoar (approved):

WalterMoar:
Can you connect this to the issue?

PR #550, opened 2020-02-21
Create Financing Statement view

bryan-gilbert:
To hold the FS component/form

### DIFF
--- a/ppr-ui/src/views/Home.vue
+++ b/ppr-ui/src/views/Home.vue
@@ -41,6 +41,11 @@
                       Search
                     </router-link>
                   </li>
+                  <li v-if="userCanSearch">
+                    <router-link to="financing">
+                      Financing Statement
+                    </router-link>
+                  </li>
                   <li>
                     <router-link to="about">
                       About
--- a/ppr-ui/src/views/Home.vue
+++ b/ppr-ui/src/views/Home.vue
@@ -41,7 +41,7 @@
                       Search
                     </router-link>
                   </li>
-                  <li v-if="userCanSearch">
+                  <li v-if="userCanFinanceStatement">
                     <router-link to="financing">
                       Financing Statement
                     </router-link>
@@ -87,11 +87,14 @@ export default createComponent({
     const userCanSearch = computed((): boolean => featureFlags.getFlag('search-registration-number') &&
       userIsAuthed.value)
 
+    const userCanFinanceStatement = computed((): boolean => featureFlags.getFlag('financing-statement') &&
+      userIsAuthed.value)
+
     function goSearch(): void {
       router.push('search')
     }
 
-    return { goSearch, userCanSearch, userIsAuthed }
+    return { goSearch, userCanSearch, userCanFinanceStatement, userIsAuthed }
   }
 })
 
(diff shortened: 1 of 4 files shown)


ppr-ui/src/views/Home.vue
```diff
@@ -41,6 +41,11 @@
                       Search
                     </router-link>
                   </li>
+                  <li v-if="userCanSearch">
```
WalterMoar: Apologies that the Issue wasn't clear - this should be under a NEW feature flag, since this is not search.

ppr-ui/tests/unit/FinancingStatement.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    wrapper = shallowMount(FinancingStatement, { })
+  })
+
+  afterEach((): void => {
+    wrapper.destroy()
+  })
+
+  it('Test the financing view', (): void => {
+    expect(wrapper.get('#financingStatement').text()).toContain('Financing Statement')
+  })
```
WalterMoar, line 23: Safe to assume this is a placeholder test? I'm not sure it's is really testing any functionality, since the component doesn't have any yet...

ppr-ui/tests/unit/FinancingStatement.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+import FinancingStatement from '@/views/FinancingStatement.vue'
+
+Vue.use(Vuetify)
+Vue.use(VueCompositionApi)
+
+describe('FinancingStatement.vue', (): void => {
+  let wrapper: Wrapper<Vue>
+
+  beforeEach((): void => {
+    wrapper = shallowMount(FinancingStatement, { })
```
WalterMoar, line 14: Can the empty braces be removed?

WalterMoar (changes requested):
Looks good, but needs its own feature flag.

ppr-ui/src/views/Home.vue
bryan-gilbert: Assumed that all new work will be released with one market feature.  Ought we see if the business needs to manage these as separate features?

ppr-ui/tests/unit/FinancingStatement.spec.ts
bryan-gilbert: It gives 100% coverage, ;-)

WalterMoar: Yes, there is that :)

ppr-ui/src/views/Home.vue
WalterMoar: No, they're separate features. We can turn both on and off at the same time if that is what the business wants, but if they're combined then we lose the ability to do one or the other, or to open them up to different audiences.

bryan-gilbert: ok thanks.

ppr-ui/tests/unit/FinancingStatement.spec.ts
bryan-gilbert: no. the tell the component you have empty options.  Without this the tests complain: 
[Vue warn]: Unknown custom element: <router-link> - did you register the component correctly? For recursive components, make sure to provide the "name" option.
      
      found in
      
      ---> <Anonymous>
             <Root>

bryan-gilbert:
@WalterMoar feature flag in place

WalterMoar (approved):

PR #562, opened 2020-02-24
Registering party components

WalterMoar:
Added the Registering Party component, with a base component that represents a Person.


ppr-ui/src/components/BaseParty.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <v-card flat>
+    <v-form @input="formValid($event)">
+      <v-text-field
+        data-test-id="BaseParty.firstName"
```
WalterMoar, line 5: As discussed with Roland, a unique "data-test-id" attribute gives as an unfragile way of locating a DOM element. This attribute name is what the other teams are using.

bryan-gilbert (approved):

PR #570, opened 2020-02-25
Changed Person to immutable PersonModel

WalterMoar:
Code reviews and discussions have highlighted the following improvements that should be made to the Person class:

- Rename it to PersonModel
- It should be immutable (remove getters)
- The constructor arguments should have default values of ''


bryan-gilbert (approved):
Interesting how the rename turned into a delete and new rather than a move and edit.  Not a problem for this very new file yet something to watch for when the file has more development history.

PR #577, opened 2020-02-26
JSONifying of the Person

WalterMoar:
Added the to/from JSON methods for PersonModel

### DIFF
--- a/ppr-ui/src/components/person-model.ts
+++ b/ppr-ui/src/components/person-model.ts
@@ -44,11 +44,11 @@ export class PersonModel {
    * Gets the JSON string representation of the PersonModel object.
    */
   public toJson(): string {
-    return '{' +
-      '"firstName":"' + this.firstName + '",' +
-      '"middleName":"' + this.middleName + '",' +
-      '"lastName":"' + this.lastName + '"' +
-      '}'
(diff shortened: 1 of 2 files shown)


ppr-ui/src/components/person-model.ts
```diff
@@ -39,4 +39,30 @@ export class PersonModel {
...
+
+  /**
+   * Gets the JSON string representation of the PersonModel object.
+   */
+  public toJson(): string {
+    return '{' +
+      '"firstName":"' + this.firstName + '",' +
+      '"middleName":"' + this.middleName + '",' +
+      '"lastName":"' + this.lastName + '"' +
+      '}'
```
bryan-gilbert: what about 
```
    return `{ "firstName": "${this.firstName}", "middleName" : "${this.middleName}", "lastName" : "${this.lastName}"}`

bryan-gilbert (approved):

ppr-ui/src/components/person-model.ts
WalterMoar: Niiiice.

WalterMoar: Will keep it on separate lines for when we add the other fields.

PR #581, opened 2020-02-26
Add form to submit and handle display of registered financing statement

bryan-gilbert:
Force page refresh if page url changes
Enums are singular
Fix our temporary load indicator


bryan-gilbert:
need to fix things .. back in a few

PR #608, opened 2020-02-27
Add form to submit and handle display of registered financing statement

bryan-gilbert:
Force page refresh if page url changes
Enums are singular
Fix our temporary load indicator
Unit test the FS view 

Had to move the Vue.use(vueRouter) out to main.ts.
Had to then adjust other page tests.

### DIFF
--- a/ppr-ui/src/assets/styles/styles.scss
+++ b/ppr-ui/src/assets/styles/styles.scss
@@ -18,4 +18,3 @@ and mixins there is no CSS code bloat.
 .app-body {
   padding-bottom: 2rem;
 }
-  
--- a/ppr-ui/src/financing-statement/FinancingStatement.vue
+++ b/ppr-ui/src/financing-statement/FinancingStatement.vue
@@ -1,55 +1,43 @@
 <template>
   <v-card outlined>
-    <form-section-header label="Type &amp; Duration" />
-    <v-container>
-      <v-form
-        v-if="editing"
-        @input="formValid($event)"
-      >
-        <v-select
-          :value="value.type"
-          :items="fsTypes"
-          label="Type"
-          name="typeInput"
-          @input="updateType"
-        />
-        <v-text-field
-          :value="value.life"
-          :rules="lifeRules"
-          label="Life in Years"
-          name="lifeInput"
-          @input="updateTerm"
-        />
-      </v-form>
-      <div v-else>
-        <div>
-          Type: {{ value.type }}
+    <v-form @input="formValid($event)">
+      <form-section-header label="Type &amp; Duration" />
+      <v-container>
+        <div v-if="editing">
+          <v-select
+            :value="value.type"
+            :items="fsTypes"
+            label="Type"
+            name="typeInput"
+            @input="updateType"
+          />
+          <v-text-field
+            :value="value.life"
+            :rules="lifeRules"
+            label="Life in Years"
+            name="lifeInput"
+            @input="updateLife"
+          />
         </div>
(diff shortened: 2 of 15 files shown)


ppr-ui/src/assets/styles/styles.scss
```diff
@@ -18,4 +18,3 @@ and mixins there is no CSS code bloat.
 .app-body {
   padding-bottom: 2rem;
 }
-
```
WalterMoar, line 21: One of those GitHub complaints again. After we get the strict TypeScript done, we should look at that issue for more opinionated formatting of .vue files.

ppr-ui/src/financing-statement/FinancingStatement.vue
```diff
@@ -32,20 +32,31 @@
     </v-container>
     <form-section-header label="Registering Party" />
     <v-container>
-      <registering-party
-        :value="value.registeringParty"
-        :editing="editing"
-        @input="updateRegisteringParty"
-        @valid="formValid($event)"
-      />
+      <v-form
```
WalterMoar: I'll need to play with this to understand why we need another form - the editing=true RegisteringParty already has a form. Also I'm not sure that we want to have two different RegisteringParty elements (one editable, the other not) if we can just have one.

My concern is getting this as simple as possible, otherwise the pattern is going to be duplicated everywhere.

ppr-ui/src/load-indicator/LoadIndicator.vue
```diff
@@ -31,3 +34,26 @@ export default createComponent({
...
+  background-color: rgba(0, 0, 0, 0.25);
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  z-index: 9999;
+}
+
+.loading__content {
+  // margin: 0 auto;
```
WalterMoar: This should be removed.

WalterMoar (changes requested):
A couple of minor changes requested, and one major question about the additional form and duplicated component.

ppr-ui/src/load-indicator/LoadIndicator.vue
```diff
@@ -31,3 +34,26 @@ export default createComponent({
 })
 
 </script>
+
+<style lang="scss" scoped>
+.loading-container {
+  position: absolute;
+  // position: relative;
```
WalterMoar: This should be removed.

ppr-ui/src/load-indicator/LoadIndicator.vue
```diff
@@ -5,10 +5,13 @@
       v-show="isLoading"
       class="loading-container"
     >
-      <div class="loading__content">
+      <div
+        align="center"
```
WalterMoar: This was identified as a code smell.

ppr-ui/src/assets/styles/styles.scss
bryan-gilbert: this is so strange ... there is a newline at the end of the file. This new line did have a space but now it doesn't. So why is github complaining?

WalterMoar: I can see why complaining about a space, since it's looking for a bare newline as the last line. Should be OK now, then?

ppr-ui/src/financing-statement/FinancingStatement.vue
bryan-gilbert: agreed

WalterMoar (approved):

PR #616, opened 2020-02-27
Added post and get calls for financing statements

WalterMoar:


ppr-ui/src/financing-statement/financing-statement-model.ts
```diff
@@ -46,11 +57,25 @@ export class FinancingStatementModel {
...
+   */
+  public toJson(): FinancingStatementInterface {
+    return {
+      securedParties: [],
+      debtors: [],
+      vehicleCollateral: [],
+      generalCollateral: [],
+      type: this.type,
+      years: this.life,
+      registeringParty: this.registeringParty.toJson()
```
bryan-gilbert, line 71: property order --- we can sort this later?

jguertin (approved):

ppr-ui/src/views/FinancingStatementView.vue
```diff
@@ -72,21 +74,27 @@ export default createComponent({
...
     }
 
     if (regNum) {
       loadIndicator.start()
-      // replace this delay with the API GET
-      setTimeout(() => {
-        financingStatement.value = new FinancingStatementModel(undefined, 25, new PersonModel('John', 'M', 'Doe'))
+
+      const url = Config.apiUrl + 'financing-statements/' + regNum
+      axiosAuth.get<FinancingStatementInterface>(url).then((response): void => {
```
bryan-gilbert, line 91: We'll need to update the unit tests to use mock axios calls and remove settimeouts ... can do this later too during our overall review.

bryan-gilbert (approved):

ppr-ui/src/financing-statement/financing-statement-model.ts
WalterMoar: I just went with what was in the API, but good call.

PR #626, opened 2020-03-02
Typescript strict fixes Issue 291

bryan-gilbert:
Set the typescript flag to 'true', resolve issues, and enhance unit test on our temporary load indicator

### DIFF
--- a/ppr-ui/src/utils/auth-helper.ts
+++ b/ppr-ui/src/utils/auth-helper.ts
@@ -5,14 +5,17 @@ import Config from '@/utils/Config'
 export function authRedirect(): void {
   if (!sessionStorage.getItem('KEYCLOAK_TOKEN')) {
     const authUrl = Config.authApiUrl
-    console.debug('auth redirect to authUrl', authUrl)
     window.location.href = authUrl
   }
 }
 
 export function getJwtValue(key: string): string {
   const jwt = sessionStorage.getItem('KEYCLOAK_TOKEN')
-  const json = jwt ? JwtDecode(jwt) : {}
-
-  return json[key]
+  if (jwt) {
+    interface Dict { [key: string]: string }
+    const jwtd: Dict = JwtDecode(jwt)
+    return jwtd[key]
+  }
+  return ''
 }
+
--- a/ppr-ui/tests/unit/LoadIndicator.spec.ts
+++ b/ppr-ui/tests/unit/LoadIndicator.spec.ts
@@ -31,7 +31,18 @@ describe('LoadIndicator.vue', (): void => {
     wrapper.destroy()
   })
 
-  it('Test flagging feature one', (): void => {
-    expect(vm.isLoading).toEqual(false)
+  it('Test not loading content', async (): Promise<void> => {
+    loadIndicator.stop()
+    expect(loadIndicator.isLoading()).toBeFalsy()
+    // wrapper.get('input[name="lifeInput"]').setValue('26')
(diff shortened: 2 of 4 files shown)


ppr-ui/tests/unit/LoadIndicator.spec.ts
```diff
@@ -31,7 +31,18 @@ describe('LoadIndicator.vue', (): void => {
     wrapper.destroy()
   })
 
-  it('Test flagging feature one', (): void => {
-    expect(vm.isLoading).toEqual(false)
+  it('Test not loading content', async (): Promise<void> => {
+    loadIndicator.stop()
+    expect(loadIndicator.isLoading()).toBeFalsy()
+    // wrapper.get('input[name="lifeInput"]').setValue('26')
```
WalterMoar, line 37: This looks like it can be removed.

ppr-ui/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "esnext",
     "module": "esnext",
-    "strict": false,
+    "strict": true,
```
WalterMoar, line 5: Hallelujah.

ppr-ui/src/utils/auth-helper.ts
```diff
@@ -5,14 +5,17 @@ import Config from '@/utils/Config'
...
   }
 }
 
 export function getJwtValue(key: string): string {
   const jwt = sessionStorage.getItem('KEYCLOAK_TOKEN')
-  const json = jwt ? JwtDecode(jwt) : {}
-
-  return json[key]
+  if (jwt) {
+    interface Dict { [key: string]: string }
```
WalterMoar, line 15: This isn't always true, BTW. But we can change it when we start dealing with the RBAC.

ppr-ui/tests/unit/LoadIndicator.spec.ts
```diff
@@ -31,7 +31,18 @@ describe('LoadIndicator.vue', (): void => {
...
+    await Vue.nextTick()
+    expect(wrapper.find('.loading__content').exists()).toBe(false)
   })
+
+  it('Test loading content', async (): Promise<void> => {
```
WalterMoar, line 42: Should we just define every test as async, whether we need it or not?

WalterMoar (approved):

ppr-ui/tests/unit/LoadIndicator.spec.ts
bryan-gilbert: nice

ppr-ui/src/utils/auth-helper.ts
bryan-gilbert: do we need a comment here?

ppr-ui/tests/unit/LoadIndicator.spec.ts
bryan-gilbert: .. because ?

WalterMoar: Sorry - the commented out line 37 isn't used for anything? (I wish GitHub would only show the line I select).

ppr-ui/src/utils/auth-helper.ts
WalterMoar: Nope - it will just get more complicated when we're returning more than strings. Just an FYI that like everything it's a work in progress.

PR #635, opened 2020-03-03
Temporarily removed payment integration until account based payment is supported

jguertin:
We've been using a hard-coded payload to payments based on the legal api to integrate with payments.  This is no longer working as we are getting back a FORBIDDEN response indicating that the user does not have permission to use the resource.

The relationships team is working on improvements to the Payment API that will better support our needs and leverage the user's account.

Rather than trying to further shoe-horn the current state to work with our system, I've temporarily commented out the integration with the payments API.


jguertin:
Closing, as we'll take a different approach

PR #674, opened 2020-03-04
Prod route needs to include WWW

bryan-gilbert:


ppr-ui/openshift/ppr-ui-dc.prod.param
```diff
@@ -17,5 +17,5 @@ ENVIRONMENT="prod"
 IMAGE_TAG="prod"
 PAY_API_URL="https://pay-api.pathfinder.gov.bc.ca/api/v1"
 PPR_GROUP="ppr-ui-prod"
-ROUTE_HOST="bcregistry.ca"
+ROUTE_HOST="www.bcregistry.ca"
```
WalterMoar, line 20: And we don't need both routes, like the way coops did it? (could be a mistake on their part). We just need to make sure both URLs will eventually work.

bryan-gilbert: This commit is to get the prod working and it will work with the www route and not with the simple hostname.

bryan-gilbert: Suggest we create a tech debt to add the second route It's particular to prod and requires a prod-only configuration file.   While this commit means our DC scripts are sufficient to deploy to prod.

WalterMoar (approved):

bryan-gilbert:
#681 Allow access via a URL without WWW

PR #688, opened 2020-03-05
add years validation fs

gh2os:
Add the validation of years field on Financing Statement API

### DIFF
--- a/ppr-api/src/schemas/financing_statement.py
+++ b/ppr-api/src/schemas/financing_statement.py
@@ -25,7 +25,7 @@ class FinancingStatementBase(pydantic.BaseModel):
     def validate_years(cls, years):  # pylint:disable=no-self-argument # noqa: N805
         if not type(years) is int:
             raise TypeError('Only integers are allowed')
-        if years < 0 or years > 25:
+        if years <= 0 or years > 25:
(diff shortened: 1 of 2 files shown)


ppr-api/src/schemas/financing_statement.py
```diff
@@ -21,6 +21,15 @@ class FinancingStatementBase(BaseModel):
     vehicleCollateral: typing.List[dict]
     generalCollateral: typing.List[dict]
 
+    @pydantic.validator('years', pre=True)
+    def validate_years(cls, years):  # pylint:disable=no-self-argument # noqa: N805
+        if not type(years) is int:
+            raise TypeError('Only integers are allowed')
+        if years < 0 or years > 25:
+            raise ValueError('Years can only be between 1-25.')
```
WalterMoar: Probably want to also include tests for the boundaries: 0, 1, and 26 (25 is already tested). Can't hurt to do some fractional values too.

gh2os: Okay, sounds good. 

Also, i noticed i should have years be `<= 0` and not `< 0` , so that will get changed.

jguertin (approved):

WalterMoar (approved):

PR #692, opened 2020-03-05
686 alembic add birthdate to fs party

gh2os:
Alembic migration to add `birthdate` field to `fs_party` table


bryan-gilbert (approved):

ppr-api/migrations/versions/324c3ef5f86c_add_birthdate_field_to_financing_.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '324c3ef5f86c'
+down_revision = 'b80c93987478'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('fs_party', sa.Column('birthdate', sa.DateTime, nullable=True))
```
jguertin: Please use date instead of DateTime

jguertin (changes requested):

jguertin (approved):

PR #743, opened 2020-03-11
Add secured party list to financing statement

bryan-gilbert:
Also needed to adjust person model to implement interface to implement fromJson

### DIFF
--- a/ppr-ui/src/components/person-name-model.ts
+++ b/ppr-ui/src/components/person-name-model.ts
@@ -1,22 +1,25 @@
 
+interface PersonNameInnerInterface {
+  first: string;
+  middle: string;
+  last: string;
+}
 /**
  * The interface to a person name.
  */
 export interface PersonNameInterface {
-  personName: {
-    first: string;
-    middle: string;
-    last: string;
-  };
+  personName: PersonNameInnerInterface;
 }
 
 /**
  * The model for a person name, such as for a registering party.
  */
-export class PersonNameModel {
-  private _first: string
-  private _middle: string
-  private _last: string
+export class PersonNameModel implements PersonNameInterface {
+  private _personName: PersonNameInnerInterface = {
+    first: '',
+    middle: '',
+    last: ''
+  }
 
   /**
    * Creates a new Person Name model instance.
@@ -26,30 +29,37 @@ export class PersonNameModel {
    * @param last the last name of the person.
    */
   public constructor(first: string = '', middle: string = '', last: string = '') {
-    this._first = first
-    this._middle = middle
-    this._last = last
+    this._personName.first = first
+    this._personName.middle = middle
+    this._personName.last = last
+  }
+
+  /**
+   * Gets personName
+   */
+  public get personName(): PersonNameInnerInterface {
+    return this._personName
   }
 
   /**
    * Gets the first name of the person.
    */
   public get first(): string {
-    return this._first
+    return this._personName.first
   }
 
   /**
    * Gets the middle name of the person.
    */
   public get middle(): string {
-    return this._middle
+    return this._personName.middle
   }
 
   /**
    * Gets the last name of the person.
    */
   public get last(): string {
-    return this._last
+    return this._personName.last
   }
 
   /**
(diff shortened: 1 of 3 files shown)


jguertin (approved):

ppr-ui/src/components/person-name-model.ts
```diff
@@ -1,22 +1,25 @@
...
-export class PersonNameModel {
-  private _first: string
-  private _middle: string
-  private _last: string
+export class PersonNameModel implements PersonNameInterface {
+  private _personName: PersonNameInnerInterface = {
+    first: '',
+    middle: '',
+    last: ''
+  }
```
WalterMoar: Just curious: why create the empty object here, and then set everything in the constructor? Would it be better to combine the two?

ppr-ui/src/components/person-name-model.ts
```diff
@@ -26,30 +29,37 @@ export class PersonNameModel {
...
+    return this._personName
   }
 
   /**
    * Gets the first name of the person.
    */
   public get first(): string {
-    return this._first
+    return this._personName.first
   }
```
WalterMoar, line 51: Very convenient :)

ppr-ui/src/financing-statement/financing-statement-model.ts
```diff
@@ -139,19 +151,23 @@ export class FinancingStatementModel {
...
     let registeringParty: PersonNameModel | undefined
+    let securedParties: BasePartyModel[] | undefined
 
     if (jsonObject.registeringParty) {
-      registeringParty = new PersonNameModel(
-        jsonObject.registeringParty.personName.first,
-        jsonObject.registeringParty.personName.middle,
-        jsonObject.registeringParty.personName.last
-      )
+      registeringParty = PersonNameModel.fromJson(jsonObject.registeringParty)
```
WalterMoar, line 157: Nice!

WalterMoar (approved):

ppr-ui/src/financing-statement/financing-statement-model.ts
```diff
@@ -139,19 +151,23 @@ export class FinancingStatementModel {
...
     }
 
+    if (jsonObject.securedParties) {
+      securedParties = []
+      jsonObject.securedParties.forEach((sp: BasePartyModel): void => {
+        securedParties.push(BasePartyModel.fromJson(sp))
```
WalterMoar, line 162: FYI: strict type checking will complain about this. I am currently working with the following.

```
  public static fromJson(jsonObject: CollateralItemsInterface): CollateralItemsModel {
    const serialCollaterals: SerialCollateralModel[] = []

    if (jsonObject.serialCollateral) {
      jsonObject.serialCollateral.forEach((serialCollateral: SerialCollateralInterface): void => {
        serialCollaterals.push(SerialCollateralModel.fromJson(serialCollateral))
      })
    }

    return new CollateralItemsModel(
      serialCollaterals.length > 0 ? serialCollaterals : undefined,
      jsonObject.generalCollateral
    )
  }
```

WalterMoar: (thanks for the code to work from - saved me mega time! 😄 )

ppr-ui/src/financing-statement/financing-statement-model.ts
```diff
@@ -104,7 +116,7 @@ export class FinancingStatementModel {
       generalCollateral: [],
       registeringParty: this.registeringParty.toJson(),
       registrationDateTime: this.registrationDateTime,
-      securedParties: [],
+      securedParties: this.securedParties,
```
WalterMoar, line 119: Aside - note that this doesn't call toJson on the elements of the array. This is causing me problems elsewhere, since I have logic in the toJson of the underlying class.

ppr-ui/src/components/person-name-model.ts
bryan-gilbert: It is needed by the type system.  Without this initialization there is no inner set of properties to assign to (even though it's defined to be of type PersonNameInnerInterface.

ppr-ui/src/financing-statement/financing-statement-model.ts
bryan-gilbert: Thanks for the suggestion to improve this.

ppr-ui/src/financing-statement/financing-statement-model.ts
bryan-gilbert: I've adjusted as suggested, thanks

PR #744, opened 2020-03-11
Update financing statement reads to support secured parties in the API

jguertin:


WalterMoar (approved):

bryan-gilbert (approved):
thanks ... working on the ui side now

Issue #769, opened 2020-03-16
Task for #647 -- ppr List Item component


bryan-gilbert:
Create a Vue component that can hold any content (e.g. secured party, debtor, serial collateral).
The component can expect these properties:
- editing (true when element is showing edit form)
- index (the index number of the contents)
- list length (the length of the list of items)
Component needs to provided a name slot to insert the appropriate prompt/header for the content.
Component needs to provide a X button (located top far right corner) when the list length is greater than or equal to 2. Emit a remove event with the content's index.

When not editing hide the prompt and the remove button.

PR #770, opened 2020-03-16
Issue #769 PPR List item

bryan-gilbert:
Create a Vue component that can hold any content (e.g. secured party, debtor, serial collateral).
The component can expect these properties:
- editing (true when element is showing edit form)
- index (the index number of the contents)
- list length (the length of the list of items)

Component needs to provided a name slot to insert the appropriate prompt/header for the content.

Component needs to provide a X button (located top far right corner) when the list length is greater than or equal to 2. Emit a remove event with the content's index.

When not editing hide the prompt and the remove button.

### DIFF
--- a/ppr-ui/src/components/PprListItem.vue
+++ b/ppr-ui/src/components/PprListItem.vue
@@ -1,0 +1,99 @@
+<template>
+  <v-list-item
+    three-line
+    class="ppr-list-item"
+  >
+    <v-list-item-content>
+      <v-list-item-title
+        v-if="editing"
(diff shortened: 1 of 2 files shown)


WalterMoar (approved):
Looks good, but can we rename at some point? This will be something we want in the MHR application as well, so having PPR in the name should be avoided. Common components is a whole other topic. Also, this seems like a List rather than a List Item. (and no I have no suggestions for names but I can help ponder).

jguertin (approved):

Issue #781, opened 2020-03-17
Task for #647: Secured Parties component


bryan-gilbert:
Add SecuredParties component to display and manage a list of secured parties.
Plus
To FS view:
- use label rather than v-model
- add the model secured parties to the FS
To FS component:
- add the secured parties component
- use emitValid instead of the deprecated validForm
- store form validity for styling
- adjust update methods to account for secured parties
Remove unneeded code in BaseParty component
To base party model add a list identifier to be used by secured parties

Issue #647, opened 2020-03-03
Create SecuredPartyList component


WalterMoar:
Consists of one or more SecuredParty components, allowing the user to add and delete.

LivMac-Git:
![5+ scrollable SP and DE](https://user-images.githubusercontent.com/12178077/76243634-d8c6a400-61f5-11ea-920e-dccf5431a34b.JPG)

![Collapsed Secured Parties](https://user-images.githubusercontent.com/12178077/76243676-e845ed00-61f5-11ea-83fa-f17f0acd6b30.JPG)

Screenshots show the wireframes for the multiple debtors/secured parties using the director list style component for collapsed vs editable, and the ability to add via the "Add another Debtor" button in the bottom of the debtor group.

Note how when a partially entered or invalid debtor is collapsed it shows in an error state calling the user's attention back to it to later correct the input before submission.

Where there are more than 5 the input section becomes scrollable with the large handle/mobile-friendly style scroll bar. There is a colored area to show progress down the list as the user scrolls.

Another optional aid are the "Next Set, Previous Set" buttons which would "jump" the scrollable area to the next 5 entries in the list bringing them into view or jumping back to the last 5 before present scroll state.

PR #782, opened 2020-03-17
Issue #781; Task for #647: Secured Parties component

bryan-gilbert:
- use label rather than v-model
- add the model secured parties to the FS
To FS component:
- add the secured parties component
- use emitValid instead of the deprecated validForm
- store form validity for styling
- adjust update methods to account for secured parties
Remove unneeded code in BaseParty component
To base party model add a list identifier to be used by secured parties
Add SecuredParties component to display and manage a list of secured parties.

### DIFF
--- a/ppr-ui/src/financing-statement/FinancingStatement.vue
+++ b/ppr-ui/src/financing-statement/FinancingStatement.vue
@@ -1,6 +1,12 @@
 <template>
   <v-card outlined>
-    <v-form @input="validForm('header', $event)">
+    <v-form @input="emitValid('header', $event)">
+      <secured-parties
+        :editing="editing"
+        :value="value.securedParties"
+        @input="updateSecuredParties"
+        @valid="emitValid('securedParties', $event)"
+      />
       <form-section-header label="Type &amp; Duration" />
       <v-container>
         <div v-if="editing">
@@ -40,7 +46,7 @@
           :value="value.registeringParty"
           :editing="editing"
           @input="updateRegisteringParty"
-          @valid="validForm('registeringParty', $event)"
+          @valid="emitValid('registeringParty', $event)"
         />
       </v-container>
     </v-form>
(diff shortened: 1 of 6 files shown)


ppr-ui/src/financing-statement/FinancingStatement.vue
```diff
@@ -88,23 +100,36 @@ export default createComponent({
...
         props.value.type,
         props.value.years,
-        newPerson // props.value.registeringParty
+        newPerson, // props.value.registeringParty
+        props.value.securedParties
+      ))
+    }
+
+    function updateSecuredParties(newSecuredParties: BasePartyModel[]): void {
+      console.log('updateSecuredParties', newSecuredParties)
```
WalterMoar, line 127: Does this console.log get removed in one of the other PRs?

ppr-ui/src/financing-statement/FinancingStatement.vue
```diff
@@ -88,23 +100,36 @@ export default createComponent({
...
     }
 
     // Callback function for emitting form validity on the header section back to the parent.
-    function validForm(key: string, validElement: boolean) {
+    function emitValid(key: string, validElement: boolean) {
       validationState[key] = validElement
       const formValid = Object.values(validationState).reduce((accumulator, elementState) => {
         return accumulator && elementState
       }, true)
+      formIsValid.value = formValid
```
WalterMoar, line 113: The `formIsValid` variable is set but never seems to be used anywhere.

ppr-ui/src/financing-statement/FinancingStatement.vue
bryan-gilbert: yes

ppr-ui/src/financing-statement/FinancingStatement.vue
bryan-gilbert: Right .. it was at one moment and we might use it again if the whole form ought to show "invalid" state.  Hmm can take this out.

bryan-gilbert: I've used this in the form and that will come in a later PR (easier that updating this and the other PR).

WalterMoar (approved):
My one concern is the complexity of SecuredParties.vue. I'm wondering if the complexity can be pushed down into the ppr-list-item, so that it isn't duplicated for every list of things that we have.

However my apologies as I'm just not on the ball enough to be able to understand the whole thing at once. I think I need to noodle around with it once its merged and then try to understand it

PR #792, opened 2020-03-17
Handle parties in database with no person name

jguertin:
When a database record does not have a person name, ensure we don't try to build an empty `personName` instance when we send it back through the API

### DIFF
--- a/ppr-api/src/models/party.py
+++ b/ppr-api/src/models/party.py
@@ -27,11 +27,12 @@ class Party(BaseORM):
     address = sqlalchemy.orm.relationship('Address')
 
     def as_schema(self):
+        person_name = schemas.party.IndividualName(
+            first=self.first_name, middle=self.middle_name, last=self.last_name
+        ) if self.last_name else None
         base_args = {
             'businessName': self.business_name,
-            'personName': schemas.party.IndividualName(
-                first=self.first_name, middle=self.middle_name, last=self.last_name
-            ),
+            'personName': person_name,
             'address': self.address.as_schema() if self.address else None
         }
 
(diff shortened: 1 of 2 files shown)


ppr-api/src/models/party.py
```diff
@@ -27,11 +27,12 @@ class Party(BaseORM):
     address = sqlalchemy.orm.relationship('Address')
 
     def as_schema(self):
+        person_name = schemas.party.IndividualName(
+            first=self.first_name, middle=self.middle_name, last=self.last_name
+        ) if self.last_name else None
```
bryan-gilbert, line 32: Is the check to be both last and first name?

ppr-api/tests/unit/models/test_party.py
```diff
@@ -0,0 +1,21 @@
...
+    model = models.party.Party(business_name='Mr. Plow')
+
+    schema = model.as_schema()
+
+    assert schema.businessName == 'Mr. Plow'
+    assert schema.personName is None
+
+
+def test_party_as_schema_person_name_is_used():
+    model = models.party.Party(first_name='Homer', middle_name='Jay', last_name='Simpson')
```
bryan-gilbert, line 14: Is there another test that tries for the 9 cases?  Fxx xMx xxL FMx ...

bryan-gilbert (approved):

gh2os (approved):

ppr-api/src/models/party.py
jguertin: No, I felt that was redundant, since there will always be a first and last name if present.

ppr-api/tests/unit/models/test_party.py
jguertin: No, there's no branches there, so I didn't see much value.  What I would do, when we implement validation would be to test validation in a similar fashion.

PR #811, opened 2020-03-18
Update descriptive information for the top level of the API Specification

jguertin:


ppr-api/docs/api_contract/ppr-api.v1.yaml
```diff
@@ -168,18 +174,13 @@ paths:
...
-            type: integer
-          in: query
-          name: version
-          description: The version of the record to retrieve.  If not specified the most recent version will be retrieved.
         - schema:
             type: string
           in: header
           name: Account-Id
-          description: The current account represented by the user
+          description: The account that the user is operating on behalf of
```
bryan-gilbert: Do you want to be consistent with periods at the end of sentences?

ppr-api/docs/api_contract/ppr-api.v1.yaml
```diff
@@ -2,7 +2,7 @@ openapi: 3.0.0
 info:
   title: Personal Property Registry API
   version: '1.0'
-  description: PPR API Specs
+  description: 'The Personal Property Register (PPR) is a centralized register where legal claims to personal property can be registered, and consumers can check if there is any debt or obligation attached to goods they may wish to buy. The PPR API provides a real-time, secure, direct link between business systems and the PPR. Most PPR transactions, including searches, registrations, renewals, amendments and discharges, directly from your own software, instead of using the PPR website.'
```
bryan-gilbert: Can/should we place this somewhere on the UI side?  I'm thinking in the main html file.

bryan-gilbert (approved):
Nice!
Trust you'll take appropriate action on my comments.

ppr-api/docs/api_contract/ppr-api.v1.yaml
jguertin: Fixed

ppr-api/docs/api_contract/ppr-api.v1.yaml
jguertin: Out of scope here, makes sense to me, but I would think someone is responsible for that content.

PR #854, opened 2020-03-20
Issue 852 (task of 823) Don't show invalid form indicators on readonly

bryan-gilbert:
When showing the successful registration the page had the red boxes.


jguertin (approved):

WalterMoar (approved):
Nice, but if we can minimize future changes to the base party it would be appreciated.

PR #859, opened 2020-03-20
Add Validation to Name Fields

gh2os:
First past at adding some validation to the Party Schema:

- personName is alphabetical other than hyphen ( - ) or apostrophe ( ' )

- businessName is alphanumeric plus allows the following special characters: & ' : , $ - ( ) % . + " " ; /

- Party cannot have businessName and personName

Also added basic tests for IndividualName and Address classes

### DIFF
--- a/ppr-api/src/schemas/party.py
+++ b/ppr-api/src/schemas/party.py
@@ -1,5 +1,6 @@
 import datetime
 import enum
+import re
 
 import pydantic
 
@@ -15,6 +16,11 @@ class IndividualName(pydantic.BaseModel):
     middle: str = None
     last: str
 
+    @pydantic.validator('*')
+    def validate_name_fields(cls, v):  # pylint:disable=no-self-argument # noqa: N805
+        assert re.match("^[a-zA-Z-']*$", v), '"IndividualName" contains invalid character'
+        return v
+
 
 class Address(pydantic.BaseModel):
     street: str
@@ -29,6 +35,13 @@ class Party(pydantic.BaseModel):
     personName: IndividualName = None
     businessName: str = None
     address: Address = None
+
+    @pydantic.validator('businessName')
+    def validate_alphanum_special(cls, v, values):  # pylint:disable=no-self-argument # noqa: N805
+        if values['personName'] is not None:
+            raise ValueError('Cannot have "personName" and "businessName"')
+        assert re.match('^[a-zA-Z-0-9-\'"&:,$%.+;/ ]*$', v), '"businessName" contains invalid character'
+        return v
 
 
 class Debtor(Party):
--- a/ppr-api/src/schemas/party.py
+++ b/ppr-api/src/schemas/party.py
@@ -41,7 +41,7 @@ class Party(pydantic.BaseModel):
         if values['personName'] is not None:
             raise ValueError('Cannot have "personName" and "businessName"')
         assert re.match('^[a-zA-Z-0-9\'"&:,$%.+;/ ]*$',
-                        v), '"businessName" can only contain the following special characters: & \' : , $ - ( ) % . + "  ; /'
+                        v), '"businessName" can only contain the following special characters: &\':,$-()%.+";/ '
         return v
 
 
--- a/ppr-api/src/schemas/party.py
+++ b/ppr-api/src/schemas/party.py
@@ -18,7 +18,7 @@ class IndividualName(pydantic.BaseModel):
 
     @pydantic.validator('*')
     def validate_name_fields(cls, v):  # pylint:disable=no-self-argument # noqa: N805
-        assert re.match("^[a-zA-Z-']*$", v), '"IndividualName" contains invalid character'
+        assert re.match("^[a-zA-Z- ']*$", v), '"IndividualName" contains invalid character'
         return v
 
 
@@ -40,7 +40,8 @@ class Party(pydantic.BaseModel):
     def validate_alphanum_special(cls, v, values):  # pylint:disable=no-self-argument # noqa: N805
         if values['personName'] is not None:
             raise ValueError('Cannot have "personName" and "businessName"')
-        assert re.match('^[a-zA-Z-0-9-\'"&:,$%.+;/ ]*$', v), '"businessName" contains invalid character'
+        assert re.match('^[a-zA-Z-0-9\'"&:,$%.+;/ ]*$',
+                        v), '"businessName" can only contain the following special characters: & \' : , $ - ( ) % . + "  ; /'
         return v
 
 
--- a/ppr-api/tests/unit/schemas/test_party_schema.py
+++ b/ppr-api/tests/unit/schemas/test_party_schema.py
@@ -1,0 +1,204 @@
+import datetime
+
+import pytest
+
+import schemas.party
+
+
+def test_person_name_invalid_no_last():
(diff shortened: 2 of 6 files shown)


ppr-api/src/schemas/party.py
```diff
@@ -15,6 +16,11 @@ class IndividualName(pydantic.BaseModel):
     middle: str = None
     last: str
 
+    @pydantic.validator('*')
+    def validate_name_fields(cls, v):  # pylint:disable=no-self-argument # noqa: N805
+        assert re.match("^[a-zA-Z-']*$", v), '"IndividualName" contains invalid character'
```
WalterMoar: Does this RE allow spaces for, say, Chris De Burgh?

ppr-api/tests/unit/schemas/test_party_schema.py
```diff
@@ -0,0 +1,204 @@
...
+
+def test_person_name_valid_with_middle():
+    name = schemas.party.IndividualName(first='John', middle='Sherman', last='Doe')
+    assert name == {'first': 'John', 'middle': 'Sherman', 'last': 'Doe'}
+
+
+def test_person_name_valid_with_special():
+    name = schemas.party.IndividualName(first="J'ohn-Boy", middle="D'ean-Dre", last="D'oe-Doughington")
+    assert name == {'first': "J'ohn-Boy", 'middle': "D'ean-Dre", 'last': "D'oe-Doughington"}
+
```
bryan-gilbert: might be good to make three tests.  This will test the first name validation and not the middle nor last.  Right?

ppr-api/tests/unit/schemas/test_party_schema.py
```diff
@@ -0,0 +1,204 @@
...
+
+def test_business_name_valid_simple():
+    name = schemas.party.Party(businessName='Globotech')
+    assert name.businessName == 'Globotech'
+
+
+def test_business_name_valid_special_chars_alphanumeric():
+    name = schemas.party.Party(businessName='Aa-\'"&:,$%.+;/zZ1234567890 ')
+    assert name.businessName == 'Aa-\'"&:,$%.+;/zZ1234567890 '
```
bryan-gilbert, line 121: Good test.  Do you want to add two tests like this for first and last person name.

ppr-api/tests/unit/schemas/test_party_schema.py
```diff
@@ -0,0 +1,204 @@
...
+    except ValueError:
+        pass
+    else:
+        pytest.fail('A validation error was expected since "country" is required')
+
+
+def test_address_invalid_no_postal():
+    try:
+        schemas.party.Address(street='123 Real Street', city='Cranbrook', country='CA')
+    except ValueError:
```
bryan-gilbert, line 154: no postal .. what is address is outside Canada?

bryan-gilbert (changes requested):
Looks good.  I'm requesting some small changes. Let me know if you disagree and I'll approve.

ppr-api/src/schemas/party.py
gh2os: This one does not - businessName does, I will check with whomever is the expert on that one. I wasn't told to include spaces, but they weren't  explicit in excluding them.

ppr-api/src/schemas/party.py
```diff
@@ -30,6 +36,13 @@ class Party(pydantic.BaseModel):
     businessName: str = None
     address: Address = None
 
+    @pydantic.validator('businessName')
+    def validate_alphanum_special(cls, v, values):  # pylint:disable=no-self-argument # noqa: N805
+        if values['personName'] is not None:
+            raise ValueError('Cannot have "personName" and "businessName"')
+        assert re.match('^[a-zA-Z-0-9-\'"&:,$%.+;/ ]*$', v), '"businessName" contains invalid character'
```
jguertin: You actually have `-` in here twice.  I'm not sure about the error message, it's not quite sure if it's dealing with singular or multiple ;). In all seriousness, it might be valuable to inform the caller what characters are valid.

ppr-api/tests/unit/schemas/test_party_schema.py
gh2os: I originally had three tests as well, but I am somewhat confident (but not 100%) that the validator checks all fields ('*')? I understand it to check each individual value. I may have misunderstood something when reading up on it. Either way, I'm happy to add 2 more.

ppr-api/src/schemas/party.py
gh2os: I had it much longer before, I can never tell what's good practice. Happy to add it back in there.

ppr-api/tests/unit/schemas/test_party_schema.py
```diff
@@ -0,0 +1,204 @@
+import datetime
+
+import pytest
+
+import schemas.party
+
+
+def test_person_name_invalid_no_last():
+    try:
+        schemas.party.IndividualName(personName={'first': 'John Doe'})
```
jguertin: I don't think this is throwing a ValueError for the reason you think it should.  personName isn't a field on IndividualName.  Should be `schemas.party.IndividualName(first='John Doe')`

Same applies to the next 4 tests

ppr-api/src/schemas/party.py
gh2os: also good catch on the hyphen, thanks :)

ppr-api/tests/unit/schemas/test_party_schema.py
jguertin, line 121: Here's a question for you: should we trim whitespace from the beginning and end of a string? I don't know the answer, I'm just asking.

jguertin (changes requested):
Good work, but tests need to be corrected.  Would prefer a bit more detailed messaging, but will leave that to you.

ppr-api/src/schemas/party.py
jguertin: Longer is okay, but it can be tough as someone debugging the system to determine which characters on their input were invalid.

ppr-api/tests/unit/schemas/test_party_schema.py
gh2os: Yeah thanks for that, I originally had it testing schemas.party.Debtor(personName={} ) but changed it to the above, but not fully clearly. I think it was better off if I didn't try and push this off before the end of the day and came at it with fresh eyes :)

ppr-api/tests/unit/schemas/test_party_schema.py
jguertin: Certainly a valid point, best to get each field failing independently

ppr-api/src/schemas/party.py
```diff
@@ -15,6 +16,11 @@ class IndividualName(pydantic.BaseModel):
     middle: str = None
     last: str
 
+    @pydantic.validator('*')
+    def validate_name_fields(cls, v):  # pylint:disable=no-self-argument # noqa: N805
+        assert re.match("^[a-zA-Z- ']*$", v), '"IndividualName" contains invalid character'
```
jguertin, line 21: I should've been specific, sorry. Should probably update both error messages.

jguertin:
Closing, as it will not be completed

PR #860, opened 2020-03-20
Just documenting the technical details.

wreckit-Ralph:
I've added a lot of text in a documentation effort.


ppr-api/docs/api_contract/ppr-api.v1.yaml
```diff
@@ -541,62 +575,77 @@ components:
...
+      title: ''
       properties:
         type:
           type: string
           enum:
             - EXACT
             - SIMILAR
+          description: |-
+            Exact matches are used when searching by Serial Number, Manufactured Home Registration Number, Aircarft DOT number, or Debtor name.
+            Similar matches are used when searching by Serial Number, Manufactured Home Registration Number and Aircraft DOT number.
```
bryan-gilbert, line 588: This is a little confusing ... For example, For serial numbers is it exact or similar or both?

ppr-api/docs/api_contract/ppr-api.v1.yaml
```diff
@@ -541,62 +575,77 @@ components:
         - years
     search-result:
       type: object
+      description: Provides a link to the financing statement found after a successful search.
```
bryan-gilbert, line 578: Curious, is this a clickable hyperlink. Is this for a search that can only result in one FS? What is returned if nothing is found?

ppr-api/docs/api_contract/ppr-api.v1.yaml
```diff
@@ -497,38 +520,49 @@ components:
...
           type: string
           format: date
+          description: 'The date upon which the financing statement will expire, in YYYY-MM-DD format.'
         baseRegistrationNumber:
           type: string
+          description: 'The Base Registration Number uniquely identifies a financing statement that is registered in the PPR database. The system assigns the Base Registration Number when the information submitted for registration passes all the system edits and is accepted for registration. You will need to know this number in order to register a change statement, a renewal statement, or a discharge statement. '
         registrationDateTime:
           type: string
           format: date-time
+          description: The date when the registration was successfully created.
```
bryan-gilbert, line 565: Might it be good to mention the timezone is the registrar's time?

ppr-api/docs/api_contract/ppr-api.v1.yaml
```diff
@@ -497,38 +520,49 @@ components:
...
+                description: Model name.
               manufacturedHomeRegNumber:
                 type: string
+                description: Home registration number assigned by the Manufactured Home Registry.
             required:
               - type
               - serial
         generalCollateral:
           type: array
+          description: Contains description of collateral other than a vehicle.
```
bryan-gilbert, line 546: Should we say this is an array of string where each string can be at most 8,000 characters.  Check this is correct.

bryan-gilbert (approved):
You may take the comments as suggestions for thought.
Nice work.

jguertin:
Closing as the changes have been incorporated in #811

ppr-api/docs/api_contract/ppr-api.v1.yaml
wreckit-Ralph: Thanks for the feedback... I agree this is confusing, and I'll work on clarifying the wording a bit more.  Serial Numbers are exact, followed by similar matches

ppr-api/docs/api_contract/ppr-api.v1.yaml
wreckit-Ralph: Thank you!

PR #865, opened 2020-03-23
Added address to the party

WalterMoar:
- Added a wrapper around the legacy BaseAddress.vue component.
- Added a model for the wrapper.
- Added the model and component to the base party.

### DIFF
--- a/ppr-ui/src/base-party/BaseParty.vue
+++ b/ppr-ui/src/base-party/BaseParty.vue
@@ -3,7 +3,7 @@
     :class="getFormClass()"
     class="base-party-form"
     data-test-id="BaseParty.form"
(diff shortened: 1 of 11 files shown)


ppr-ui/src/base-party/BaseParty.vue
```diff
@@ -3,7 +3,7 @@
     :class="getFormClass()"
     class="base-party-form"
     data-test-id="BaseParty.form"
-    @input="emitValidity(HEADER, $event)"
+    @input="emitValid(HEADER, $event)"
```
bryan-gilbert, line 6: nice

bryan-gilbert (approved):

PR #866, opened 2020-03-23
Issue #834 Type of Registration

bryan-gilbert:
Was able to set a value into the select during unit tests so we've reduced our skips from 3 to 2

### DIFF
--- a/ppr-ui/src/financing-statement/FinancingStatement.vue
+++ b/ppr-ui/src/financing-statement/FinancingStatement.vue
@@ -11,6 +11,7 @@
           <v-select
             :value="value.type"
             :items="fsTypes"
+            data-test-id="FinancingStatement.type.select"
             label="Type"
             name="typeInput"
             @input="updateType"
@@ -67,7 +68,7 @@
 import { createComponent, ref } from '@vue/composition-api'
 import { BasePartyModel } from '@/base-party/base-party-model'
 import { FinancingStatementModel } from '@/financing-statement/financing-statement-model'
-import { FinancingStatementType, FinancingStatementTypeCodeList } from '@/financing-statement/financing-statement-type'
+import { FinancingStatementType } from '@/financing-statement/financing-statement-type'
 import { PersonNameModel } from '@/components/person-name-model'
 import BaseParty from '@/base-party/BaseParty.vue'
 import DebtorParties from '@/financing-statement/DebtorParties.vue'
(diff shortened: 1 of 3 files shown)


bryan-gilbert:
image

As discussed this task just adds the full range of types into the select dropdown.  Note the length of text in some of these options.

bryan-gilbert:
Note this resolves 834 not 812.

ppr-ui/src/financing-statement/financing-statement-type.ts
```diff
@@ -1,15 +1,12 @@
...
-  SALE_OF_GOODS
-  FORESTRY CONTRACTOR LIEN
-  FORESTRY CONTRACTOR CHARGE
-  FORESTRY SUBCONTRACTOR CHARGE
-  TAX LIEN
-  LAND TAX
-  NOTICE OF MARRIAGE ...
-  */
+  SECURITY_AGREEMENT = 'PPSA Security Agreement',
+  REPAIRERS_LIEN = 'Repairer\'s Lien',
```
jguertin, line 3: In Python we use `'` except where the string contains `'`, then we use `"`. Should we do the same in TS?

jguertin (approved):

ppr-ui/src/financing-statement/FinancingStatement.vue
```diff
@@ -97,7 +98,18 @@ export default createComponent({
...
+    // Create the set of text value pairs for the Type select
+    interface SelectItem {
+      text: string;
+      value: string;
+    }
+    const fsTypes: SelectItem[] = []
+    const fsTypeLabels: string[] = Object.values(FinancingStatementType)
+    Object.keys(FinancingStatementType).forEach((key: string, index: number): void => {
+      fsTypes.push({ text: fsTypeLabels[index], value: key })
+    })
```
WalterMoar, line 111: Curious - is this something that we may want outside the component, for re-use? I don't think we need it here, but possibly for the other enums.

ppr-ui/tests/unit/financing-statement/FinancingStatement.spec.ts
```diff
@@ -49,17 +49,22 @@ describe('FinancingStatmentContainer.vue', (): void => {
...
 
-    // Skip until can figure out how to inject value into a vuetify select
-    it.skip('@input - type change should be emitted', async (): Promise<void> => {
+    it('@input - type change should be emitted', async (): Promise<void> => {
       const properties = ref({ editing: true, value: new FinancingStatementModel() })
       const wrapper: Wrapper<Vue> = mount(FinancingStatement, { propsData: properties.value, vuetify })
-
-      wrapper.get('input[name="typeInput"]').setValue('REPAIRERS_LIEN')
+      const sel = wrapper.get('input[data-test-id="FinancingStatement.type.select"]')
+      sel.setValue('REPAIRERS_LIEN')
```
WalterMoar, line 56: So that's all it was?! Nice work.

WalterMoar (approved):

PR #870, opened 2020-03-26
Write detailed API specification documentation

jguertin:
First draft of an API document outlining the intent and current status.  Includes a skeleton of what functionality still needs to be defined.

### DIFF
--- a/ppr-api/docs/api_contract/README.md
+++ b/ppr-api/docs/api_contract/README.md
@@ -103,16 +103,60 @@ nor should it be available for any other write operation until payment is comple
   - Use the the `INFRG` fee code when the `lifeInfinite` property is `true`
   - Use the the `FSREG` fee code along with a quantity matching the `lifeYears` property when life isn't infinite
   - A Repairer's Lien may require its own fee code, though `FSREG` with a quantity of `1` may be sufficient
+  
+**Registering Party:** For the moment, registering party details are accepted by the API. There has been consideration
+that it would be preferable to use the logged in user as the registering party.  In that case, some decisions still need
+to be considered in how to implement that:
+- Can the user override address and name fields?
+- Should the address and name values be stored in the PPR database, or simply looked up through the systems that have
+  that information?
+- If we no longer store registering party details, how do we handle historical data from the legacy PPR system
+
+### List Financing Statements
+
+The intent of this endpoint is to provide users the ability to lookup existing financing statement records. Paging and
+filtering URL parameters have not yet been defined, but they can be added as needed.
+
+This capability did not exist in the legacy PPR system.  It could enable dashboards that give users the ability to find
+their references to their own Financing Statements without the need to perform a search.
+
+When implementing, consider limiting what portions of the financing statement are included in the payload.
+
+**Endpoint:** `GET /financing-statement`
+
+**Implementation Status:** Not implemented at all
+
+**Data Restriction:** This should only provide records that are available to the calling user.  For public users, this
+would be records associated with the provided account id.
+
+### View a Financing Statement
+
+The function is intended as a lookup capability, where a user can lookup the current details of their financing
+statement. It may be useful in dashboard management or in viewing the details or start an amendment without the need to
+perform a search.
+
+This capability did not exist for public users in the legacy PPR system. Depending on regulations, consider placing
+limitations on when data is available through this endpoint.
+
+The _financingStatementId_ URL path variable is interchangeable with base registration number.
+
+**Endpoint:** `GET /financing-statement/{financingStatementId}`
+
+**Implementation Status:** For the most part this endpoint is operational. Still outstanding:
+- Fields specific to Repairer's Liens: [Trust Indenture](https://github.com/bcgov/ppr/issues/818),
+  [Lien amount](https://github.com/bcgov/ppr/issues/819), [Surrender Date](https://github.com/bcgov/ppr/issues/820)
+- General Collateral requires some changes to support the solution outlined in this
+  [Collateral comment](https://github.com/bcgov/ppr/issues/815#issuecomment-603935842)
+- The Payment portion is not yet included in the response payload
+
+**Data Restriction:** This should only provide records that are available to the calling user.  For public users, this
+would be records associated with the provided account id. Consider whether additional restrictions apply.
 
 ### Amendments or Change Statements
 
 ### Renewal
 
 ### Discharge
-
-### View a Financing Statement
-
-### List Financing Statements
 
 ### Historical View of a Financing Statement
 
--- a/ppr-api/docs/api_contract/README.md
+++ b/ppr-api/docs/api_contract/README.md
@@ -10,7 +10,7 @@ capabilities.
 
 Both documents are a work in progress and are subject change.  Generally speaking, this API has been designed in such a
 way that write operations (`POST`, `DELETE`, etc...) correspond with capabilities that are charged, where as read
-operations (`GET`) would are geared to be lookups, and would not have a charge.
+operations (`GET`) are intended to be lookups, and would not have a charge to the end user.
 
 ### Implementation Status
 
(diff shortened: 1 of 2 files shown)


gh2os (approved):

ppr-api/docs/api_contract/README.md
```diff
@@ -0,0 +1,486 @@
...
+access to data for the web UI and will also be the tool used by external parties to interact with the Personal Property
+Registry.
+
+The [API Specification](ppr-api.v1.yaml) outlines the various RESTful endpoints that will be available in the PPR API.
+The goal of this document is to provide some context to that specification and give insight to the intended
+capabilities.
+
+Both documents are a work in progress and are subject change.  Generally speaking, this API has been designed in such a
+way that write operations (`POST`, `DELETE`, etc...) correspond with capabilities that are charged, where as read
+operations (`GET`) would are geared to be lookups, and would not have a charge.
```
bryan-gilbert: "would be geared towards lookups" ?

bryan-gilbert: what about PATCH?  Did you use PATCH in the API?

ppr-api/docs/api_contract/README.md
```diff
@@ -0,0 +1,486 @@
...
+
+## Financing Statements
+
+A financing statement can be registered once and changed or amended many times.  On the initial registration, it will be
+assigned a **Base Registration Number**, which will be used as the identifier for the financing statement. Any further
+operations must use this identifier.
+
+For purposes of the API, any write operations are considered events. Under the hood, each event references the financing
+statement.  Any relational data associated with the financing statement will have references to the starting and ending
+events. Internally, this enables us to leverage the data model to wither do a quick lookup of the current state of the
```
bryan-gilbert: "whither" ?

ppr-api/docs/api_contract/README.md
```diff
@@ -0,0 +1,486 @@
...
+financing statement or use the events to rebuild the state up to a specific point in time.
+
+When performing an operation using a `financing-statement` endpoint in the API, this leverages the current state.
```
bryan-gilbert: Can you make the meaning of this sentence a little clearer?

ppr-api/docs/api_contract/README.md
```diff
@@ -0,0 +1,486 @@
...
+
+### Amendments or Change Statements
+
+The goal of this operation is to apply a change to a financing statement. Under the BC Government statutes, there are
+several different types of amendments, each of one performs a specifically design change. There also exists a change,
+which is a more general update to financing statement that provides capabilities that equivalent to multiple amendments
+at once. Some clarification is still required on whether standalone amendments are necessary or whether having the
+ability to do changes is sufficient.
+
+As for the proposed implementation, there are a few different ways to approach this. Ultimately, we've recommended this
```
bryan-gilbert: I can't see what "this" means in line 168.  I believe the intention is to recommend just implementing "change".

ppr-api/docs/api_contract/README.md
```diff
@@ -0,0 +1,486 @@
...
+would be records associated with the provided account id.
+
+**Alternatives to Consider:**
+- An update of a financing statement could be performed with `PUT /financing-statements/{financingStatementId}`.
+  Challenges to address would be how to represent payments for individual events and how to handle general collateral
+  where pre-existing collateral is read only. A benefit is that operations such as debtor transfers could be performed
+  by reviewing the new values, removing any that weren't provided and adding any that did not previously exist. 
+- Likewise, a similar approach would be to use `PATCH /financing-statements/{financingStatementId}`. The benefit of this
+  approach is that it would not be necessary to submit all of the data in the financing statement for a change, which
+  would be a pit simpler than a PUT. Additionally, rules around handling changes would be more flexible. Challenges
```
bryan-gilbert: "pit" => "bit"

ppr-api/docs/api_contract/README.md
```diff
@@ -0,0 +1,486 @@
...
+
+**Implementation Status:** Not implemented at all
+
+**Data Restriction:** This should be allowed for records that are available to the calling user.  For public users, this
+would be records associated with the provided account id.
+
+**Alternatives to Consider:**
+- An update of a financing statement could be performed with `PUT /financing-statements/{financingStatementId}`.
+  Challenges to address would be how to represent payments for individual events and how to handle general collateral
+  where pre-existing collateral is read only. A benefit is that operations such as debtor transfers could be performed
```
bryan-gilbert, line 188: Isn't general collateral a list with append only operations? So PATCH would work well for this while PUT wouldn't

ppr-api/docs/api_contract/README.md
```diff
@@ -0,0 +1,486 @@
...
+**Implementation Status:** Not implemented at all
+
+**Data Restriction:** This should be allowed for records that are available to the calling user.  For public users, this
+would be records associated with the provided account id.
+
+**Alternatives to Consider:**
+- An endpoint specifically to create renewals could be added to the financing statement. Using
+  `POST /financing-statements/{financingStatementId}/renewals` would have the benefit of simplifying the work to be
+   explicitly for this operation, though it would add an additional endpoint to be maintained.
+- An update to life could be handled with `PATCH /financing-statements/{financingStatementId}`, thought this interface
```
bryan-gilbert: "thought" => "though"

What about PATCH /financing-statements/{fsid}/renewal

ppr-api/docs/api_contract/README.md
```diff
@@ -0,0 +1,486 @@
...
+
+### Discharge
+
+Discharge a financing statement. This should not physically delete the record, but rather create a discharge event that
+updates the status of the financing statement.
+
+Consider whether a different end point would be better suited to this operation, as the financing statement should still
+be available for read operations after this occurs.
+
+**Endpoint:** `DELETE /financing-statements/{financingStatementId}`
```
bryan-gilbert, line 233: Again, since this is yet one more event applied to the current state, and since (I believe) a discharge can be revoked within a certain time,  why not use PATCH again with the url suffix of /discharge.
Discharge is a change of state. Not a delete. Yes?

bryan-gilbert (approved):

ppr-api/docs/api_contract/README.md
jguertin: Change the wording to "(`GET`) are intended to be lookups, and would not have a charge to the end user"

As discussed, I did not specify a `PATCH` endpoint in the spec, but did mention it as a possibility in passing. `PATCH` would be counted among the write operation.

ppr-api/docs/api_contract/README.md
jguertin: "either".  Nice catch.

ppr-api/docs/api_contract/README.md
jguertin: Changed the paragraph as follows.  Still not perfect , but hopefully a little bit better.

```Operations on the `financing-statement` schema are designed as CRUD operations on a financing state.  As such, events
are abstracted out of the interface, so reading a response object can be considered the current state of the financing
statement.```

ppr-api/docs/api_contract/README.md
jguertin: Changed the first 2 paragraphs in this section as follows to provide greater clarity:

"The goal of this operation is to apply a change to a financing statement. Under the BC Government statutes, there are
several different types of amendments, each of one performs a specifically designed change. There also exists a
"change", which is a more general update to financing statement that provides capabilities that equivalent to multiple
amendments at once. Some clarification is still required on whether standalone amendments are necessary or whether
having the ability to do changes is sufficient. The recommended approach is to not implement fine grained amendments
and instead just have a "General Change" operation. 

As for the proposed implementation of changes, there are a few possible approaches. Ultimately, we've recommended a
`POST` to events as it ties changes to the event model. Some aspects of a change event to not go well with a state
based model, such as payments, extending life (renewals, described separately below) or adding general collateral."

ppr-api/docs/api_contract/README.md
jguertin: As discussed, it's not really the ideal approach, but I do mention `PATCH` as a possibility in the next bullet point.

ppr-api/docs/api_contract/README.md
jguertin: fixed 👍

ppr-api/docs/api_contract/README.md
jguertin: Fixed the typo, thanks.

There's be very little value in `PATCH` for a renewal specific endpoint, as it would have its own schema so a `POST` would be perfectly adequate should we add that sub url.

ppr-api/docs/api_contract/README.md
jguertin: It's kind of a logical delete, which partially makes sense for a delete endpoint.  It is not an uncommon approach.  Due to the complexity of the space though, neither `DELETE` nor `PATCH` is completely ideal.  That's why I've included the text "Consider whether a different end point would be better suited to this operation".  This is one of the ones where we've suggested a possible solution, but it's up to them if they decide to go that way when it comes to implementation.

PR #871, opened 2020-03-26
Financing Statement select and change type.

bryan-gilbert:
Add a confirmation dialog component.
When Repairer's Lien elements are added will need to toggle their display based on selected type.
May wish to only show the confirm change dialog after user has actually made a change.


jguertin (approved):
I certainly wouldn't mind if the coverage checks stopped wining when coverage drops, but I still like general reporting of coverage, in particular on `master`.

PR #872, opened 2020-03-26
Suppress code coverage

bryan-gilbert:


jguertin:
This seems to also disable the unit tests.  I would not recommend disabling tests.  If we break any it will make it that much more difficult to get the project up and running again.

As far as disabling the coverage reporting, I personally don't see any harm in keeping it in.  May be stale, but its still an open source repo.  Once code starts changing it's still good to have in there.  I also like it for when I point prospective clients to this repo when looking for my next gig ;).

bryan-gilbert:
Leave master with code coverage.

PR #881, opened 2020-04-01
Correct life storage in the database

jguertin:
Correct life storage to have cumulative life stored on the financing statement and individual events (base registration and renewals) stored on events.

### DIFF
--- a/ppr-api/docs/sample_data/create_sample_financing_statements.sql
+++ b/ppr-api/docs/sample_data/create_sample_financing_statements.sql
@@ -1,10 +1,11 @@
 INSERT INTO financing_statement (reg_number, reg_type_cd, status, life, expiry_date) VALUES
-    ('123456A', 'SA', 'A', 5, now() + interval '5 years'),
+    ('123456A', 'SA', 'A', 7, now() + interval '7 years'),
     ('123456B', 'SA', 'A', -1, null);
 
 INSERT INTO registration (reg_number, base_reg_number, change_type_cd, reg_date, life, document_number) VALUES
-    ('123456A', '123456A', NULL, now(), null, 'A0000001'),
-    ('123456B', '123456B', NULL, now(), null, 'A0000002'),
+    ('123456A', '123456A', NULL, now(), 5, 'A0000001'),
+    ('123456C', '123456A', NULL, now() + interval '1 hour', 2, 'A0000004'),
+    ('123456B', '123456B', NULL, now(), -1, 'A0000002'),
     ('123456Z', '123456B', 'DT', now() + interval '1 hour', null, 'A0000003');
 
 INSERT INTO vehicle (base_reg_number, reg_number_start, reg_number_end, vehicle_type_cd, mhr_number, year, make, model, serial_number) VALUES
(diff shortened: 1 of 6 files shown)


ppr-api/docs/sample_data/create_sample_financing_statements.sql
```diff
@@ -1,10 +1,11 @@
 INSERT INTO financing_statement (reg_number, reg_type_cd, status, life, expiry_date) VALUES
-    ('123456A', 'SA', 'A', 5, now() + interval '5 years'),
+    ('123456A', 'SA', 'A', 7, now() + interval '5 years'),
```
bryan-gilbert: Is this 7 intentional?

ppr-api/src/endpoints/search.py
```diff
@@ -118,6 +121,16 @@ def rebuild_financing_statement_to_event(event: models.financing_statement.Finan
...
 
+        # Apply the life of each event to capture the combined expiry. This includes renewals.
+        if not infinite:
+            if applied_event.life_in_years == -1:
+                infinite = True
+                years = None
+                expiry = None
+            elif applied_event.life_in_years and applied_event.life_in_years > 0:
+                years += applied_event.life_in_years
+                expiry += datedelta.datedelta(years=applied_event.life_in_years)
```
bryan-gilbert, line 132: How does this end up affecting expiry?  I can't see how datedelta knows about the original expiry date and can add the correct years.  Magic?

ppr-api/tests/integration/utilities/sample_data_utility.py
```diff
@@ -74,6 +76,7 @@ def create_test_financing_statement(**kwargs):
 def create_test_financing_statement_event(fin_stmt: models.financing_statement.FinancingStatement, **kwargs):
     options = dict({
         'change_type_code': None,  # A 2 character string, a RegistrationType value
+        'years': None,  # An int from 1 to 25, or -1 for Infinity
```
bryan-gilbert, line 79: Is there a naming consistency issue here?  "years" or "lifeYears"

bryan-gilbert (approved):

ppr-api/src/endpoints/search.py
jguertin: on lines 108-110 I set years to 0 and expiry to today.  So this operation increments the expiry date in the same way years is incremented the line before.

ppr-api/docs/sample_data/create_sample_financing_statements.sql
jguertin: Yes, because I added a 2 year "renewal" event.  Though I did forget to adjust `expiry_date`, so I'll fix that. thanks

ppr-api/tests/integration/utilities/sample_data_utility.py
jguertin: Yes and no.  Years was the original value and represents what's stored.  It's only used in this  test fixture this way and I opted to keep it just years for brevity.

It is consistent with the interface of the companion method `create_test_financing_statement`.

bryan-gilbert (approved):

WalterMoar (approved):

PR #884, opened 2020-04-01
Rearrange the base party and submodels

WalterMoar:
Updated the BaseParty model and the models it uses, so that fromJson handles undefined. This allows the logic to stay within the model, so that anything using the model is simpler. 

Changed the person name model to be a single layer, rather than the double set of interfaces.

Made toJson and fromJson stricter WRT handling empty objects and undefined fields.

### DIFF
--- a/ppr-ui/src/base-party/BaseParty.vue
+++ b/ppr-ui/src/base-party/BaseParty.vue
@@ -69,7 +69,8 @@ export default createComponent({
       type: String
     },
     value: {
-      required: true,
+      default: () => new BasePartyModel(),
+      required: false,
       type: BasePartyModel
     }
   },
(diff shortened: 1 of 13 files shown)


ppr-ui/src/base-party/BaseParty.vue
```diff
@@ -69,7 +69,8 @@ export default createComponent({
       type: String
     },
     value: {
-      required: true,
+      default: () => new BasePartyModel(),
+      required: false,
```
WalterMoar, line 73: Did this so that it can handle `undefined` as the model.

ppr-ui/tests/unit/components/BaseParty.spec.ts
```diff
@@ -167,6 +167,7 @@ describe('BaseParty.vue', (): void => {
       await Vue.nextTick()
       wrapper.get('input[data-test-id="BusinessName.input.name"]').setValue('NewBusinessName')
       await Vue.nextTick()
+      await Vue.nextTick()
```
jguertin, line 170: Just out of curiosity, why wait 2 ticks?

jguertin (approved):
👍

ppr-ui/tests/unit/components/BaseParty.spec.ts
WalterMoar: I would like to know! Sometimes it's the special sauce for getting a test to pass. Even though the test was fine before. I'm guessing something to do with Vue Test Utils still being beta, albeit a fairly advanced one (the next version actually breaks some tests, so I think we'll avoid that!).

PR #911, opened 2020-04-06
Fixed bug where undefined used in JSON

WalterMoar:
The TypeScript value `undefined` was erroneously being used in JSON. Fixed this for the person name, as it was very visible for missing middle names. It likely needs to be done elsewhere, TBD.

### DIFF
--- a/ppr-ui/src/components/person-name-model.ts
+++ b/ppr-ui/src/components/person-name-model.ts
@@ -53,11 +53,21 @@ export class PersonNameModel {
    * Gets the JSON representation of the PersonNameModel object.
    */
   public toJson(): PersonNameInterface {
-    return {
-      first: this.first,
-      middle: this.middle,
-      last: this.last
+    let rval: PersonNameInterface = {}
+
+    if (this.first) {
+      rval = Object.assign(rval, { first: this.first })
     }
+
+    if (this.middle) {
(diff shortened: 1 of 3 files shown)


ppr-ui/src/components/person-name-model.ts
```diff
@@ -53,11 +53,21 @@ export class PersonNameModel {
...
-    return {
-      first: this.first,
-      middle: this.middle,
-      last: this.last
+    let rval: PersonNameInterface = {}
+
+    if (this.first) {
+      rval = Object.assign(rval, { first: this.first })
+    }
+
```
WalterMoar, line 61: Is there a better way to do this in ES/TypeScript?

jguertin: Another approach might be to use nulls:

```
return {
  first: this.first || null,
  middle: this.middle || null,
  last: this.last || null,
}
```

But if you want to omit the undefined values I think what you have is the way to go.

jguertin (approved):

WalterMoar:
@jguertin That might be the special sauce needed to do it - define as null in the interface and use TypeScript's null/undefined "sorta equivalence" to bridge the gap.

(errr, nevermind, I turned on strict typechecking and that won't fly. Oh well).

PR #913, opened 2020-04-07
Improve linting with flake8-docstrings and pydocstyle

jguertin:
Added additional linting to get the code style more in line with other projects.  In doing this, I also filled the API with quite a bit more documentation.

It is by no means comprehensive, as I was short on time, so my goal here was to  add a one liner where needed just to help document the intent of the system.  This should at least provide some more insight going forward.

### DIFF
--- a/ppr-api/src/auth/authentication.py
+++ b/ppr-api/src/auth/authentication.py
@@ -1,3 +1,5 @@
+"""A module that provides methods for accessing the Auth API and providing the logged in user details."""
+
 import http
 import json
 import logging
@@ -17,6 +19,7 @@ bearer_scheme = fastapi.security.HTTPBearer()
 
 
 def check_auth_response(response: requests.Response):
+    """Review the response from the external response and throw an error if it was forbidden or unauthorized."""
     if response.status_code in [http.HTTPStatus.UNAUTHORIZED, http.HTTPStatus.FORBIDDEN]:
         try:
             body = response.json()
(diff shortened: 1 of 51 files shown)


ppr-api/src/auth/authentication.py
```diff
@@ -17,6 +19,7 @@
 
 
 def check_auth_response(response: requests.Response):
+    """Review the response from the external response and throw an error if it was forbidden or unauthorized."""
```
WalterMoar, line 22: Response from the external response?

jguertin: Like, I said just trying to get something is better than nothing ;) Will fix on my next PR if that works for you.

"Review the response from the external API"

ppr-api/src/models/financing_statement.py
```diff
@@ -67,19 +72,25 @@ def as_schema(self):
...
     def get_base_event(self):
+        """Find the event created for the Base Registration."""
         return next((e for e in self.events if e.registration_number == self.registration_number), None)
 
     def get_debtors(self):
+        """Filter the parties to get only the debtors."""
         return list(filter(lambda p: p.type_code == PartyType.DEBTOR.value, self.parties))
 
     def get_registering_party(self):
+        """Fine the registering party from the list of active parties."""
```
WalterMoar, line 83: Find?

jguertin: Yes. Spellcheck doesn't catch these for me.... lol

WalterMoar (approved):

